PR #7, opened 2015-10-30
changed to resemble more a pharo build process

estebanlm:
... but I cannot tests it. 
I hope it works :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,14 +53,16 @@ env:
 language: smalltalk
 sudo: false
 smalltalk:
+  - Pharo-latest
+  - Pharo-stable
   - Pharo-5.0
   - Pharo-4.0
-  - Pharo-3.0
 env:
   global:
     - BASELINE=myProject
-    # - BASELINE_GROUP="TravisCI" # Name of the group to load from baseline
-    # - PACKAGES="/packages" # Directory where Filetree looks for packages
+    # - TESTS=PackagesToTest      # RegEx, i.e.: [A-Z].*, default is defined as BASELINE.*
+    # - BASELINE_GROUP="default"  # Name of the group to load from baseline
+    # - PACKAGES="/packages"      # Directory where Filetree looks for package
 ```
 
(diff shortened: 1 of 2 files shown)


fniephaus:
Hi @estebanlm,
The code you removed was meant to enable caching when the script runs locally.
Do you really want to get rid of this for Pharo? If a build fails, filetreeCI prints a "copy-and-pastable" command to reproduce the build locally for debugging purposes.

Also, why did you remove `Pharo-3.0`? Does it harm to have it?

fniephaus:
I have merged your changes into [dev](https://github.com/hpi-swa/filetreeCI/tree/dev) and re-add Pharo caching and `Pharo-3.0` (e7fd6ff36729bdbcb20201e1b64c3f7cf3d4c4cf).

Here's a test build: https://travis-ci.org/hpi-swa/filetreeCI/builds/88453271

What do you think?

estebanlm:
that will not work (script for vm’s wont work). 
I do not have time today but monday I will propose a different solution. 

why we need caching? I do not want to cache the vm, is not a problem for us to download a new one each time.

fniephaus:
What will not work? `Pharo-3.0`? It seems to be working already (or am I missing something?).
https://travis-ci.org/hpi-swa/filetreeCI/jobs/88453288

Feel free to propose more changes to [dev](https://github.com/hpi-swa/filetreeCI/tree/dev), as we can run the tests before pushing to master.

The caching is meant for debugging purposes when running filetreeCI locally. It might not be as useful to Pharo users, since Pharo already comes with built-in command line test capabilities, but it makes local end-to-end tests less painful.
Instead of running a script, that runs two scripts to download each, a vm and an image, the code currently directly runs the two scripts (e.g. `/vm50` + `/50`) and saves the output to a cache folder.
Here's what it looks like on my computer at the moment:
image
IMHO it'd be nice to keep it, but if you really want to get rid of it, we can do that as well.

PR #14, opened 2015-11-30
repository need to include "/" so you can specify local files as "." …

estebanlm:
…or "name" (otherwise you need to add "/." and "/name" which is incorrect.

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -84,7 +84,7 @@ print_info "Loading project..."
 $PHARO_VM "$SMALLTALK_CI_BUILD/$PHARO_IMAGE" eval --save "
 Metacello new 
     baseline: '${BASELINE}';
-    repository: 'filetree://${PROJECT_HOME}${PACKAGES}';
+    repository: 'filetree://${PROJECT_HOME}/${PACKAGES}';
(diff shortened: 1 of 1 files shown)


fniephaus:
What's the problem with the current definition?
Doesn't `/.` work as intended? You just want to get rid of the very first slash?

fniephaus:
Maybe the README.md is misleading here again and might only need another update?
I just want it to be as consistent as possible across all dialects and I decided to use the slash to make clear its the user's repositories root, if you know what I mean.

estebanlm:
yes, but the slash is non-consistent with the rest of the world, IMO. 
the (IMO, correct) assumption is relative, not absolute... also "root of the project" means in general "from here to down, and slash means "from root of the disk to down"... 
Not that I cannot live with a clarification in the README, but I think this is not correct in general. 

btw... I have a couple of questions related to caching, etc... where should I send them? :P

marceltaeumel:
Yes, using "/something" for a local path is strange. That slash should be part of the outer scope.

fniephaus:
I will update this later today. Thanks for the feedback.
@estebanlm feel free to send your questions to code@fniephaus.com :smiley:

PR #15, opened 2015-12-06
Running tests now produce junit xml artifacts.

peteruhnak:
The default test output is not particularly great, partially because Pharo uses CR as line separators, so the output tends to overlap.

However the test runner can produce junit xml artifact... pretty much a universal output across languages and tools.

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -89,5 +89,5 @@ Metacello new
 "
 
 print_info "Run tests..."
-$PHARO_VM "$SMALLTALK_CI_BUILD/$PHARO_IMAGE" test --fail-on-failure "$TESTS" 2>&1 || EXIT_STATUS=$?
+$PHARO_VM "$SMALLTALK_CI_BUILD/$PHARO_IMAGE" test --junit-xml-output --fail-on-failure "$TESTS" 2>&1 || EXIT_STATUS=$?
 # ==============================================================================


fniephaus:
Thanks for the PR, this seems reasonable.
Does it make sense to add a more generic env variable to define custom test flags (e.g. `$PHARO_TEST_FLAGS` with `--junit-xml-output --fail-on-failure` as default).
What do you and @estebanlm think? 

Also, this could be interesting for @bahnfahren, @chistopher, @jchromik and @Nef10. AFAIK they have been working on some JUnit/XML related stuff.

peteruhnak:
All the test options are:

```
Usage: test [--junit-xml-output] [--fail-on-failure] [<package> ...]
	--junit-xml-output    output the test results in a junit compatible format
	--fail-on-failure     if there is a test error or failure, it will exit with error code 1
	--save                save after executing tests
	 <package>            a String matching a package name
```

So maybe an env would make sense, although I do not like the idea of having X global variables (or global variables at all).

Would it make sense to have more semantic/hierarchical config?

e.g.
```
project:
    - baseline: 'BaselineOfMyProject'
    - group: 'default'
    - directory: '/repository'

tests:
    - packages: "MyProject-*"
    - flags: [ fail-on-failure, junit-xml-output ]
```

fniephaus:
Ah ok, so there are only three flags anyway and `--junit-xml-output` seems to be the only flag that one might want to toggle.
I'm also not a big fan of too many global variables, but this is the only way to configure builds on Travis.
I will merge this for now...if there are any objections, we can still make this configurable later.

We are currently waiting for Travis to merge a couple of PRs [[1](https://github.com/travis-ci/travis-build/pull/546), [2](https://github.com/travis-ci/docs-travis-ci-com/pull/419)] and then the config will look something like this:
```
language: smalltalk
smalltalk:
  - Squeak-5.0
  - Squeak-trunk
baseline:
  - MyBaseline
```

This project is still very new, so we are happy to incorporate more feedback and make it better.
If you have further suggestions/ideas, please feel free to open another issue.

estebanlm:
sorry for coming late, I was not around the weekend. 
first, yes I would be ok with a PHARO_TEST_FLAGS with a default as described. 
second, I'm also not a fan of environment flags and I would prefer a better described config, as suggested by travis in general (I already talked with @fniephaus about... I can live with env variables is there is a goof reason, of course).

fniephaus:
@peteruhnak now that `--junit-xml-output` is added, what should happen to the `.xml` file once it is created after a build? Print to stdout? Unfortunately, Travis doesn't have a built-in/free build artifacts upload mechanism.

chistopher:
We're experimenting with [printing $SMALLTALK_CI_HOME/result.xml](https://github.com/HPI-BP2015H/travis-build/blob/xml-output/lib/travis/build/script/smalltalk.rb#L53) because travis saves logs of all output. Nothing official, though.

fniephaus:
Pharo creates a `xml` output file (e.g. `SWTDemo-Tests-Test.xml`) within the build folder.
So this should work to print all files:

```bash
cat "${SMALLTALK_CI_BUILD}"/*.xml
```

I honestly haven't fully understood what to do with these files when they are in the Travis log. Sure, they can be parsed out, but then? Why would I want to view them in a JUnit XML viewer or create a pdf report? Or what else can I do with them?

peteruhnak:
The main reason why I wanted the XML was because the default output is not very useful (partially because linux uses LF as line separators and Pharo CR, so the lines overlap each other).

So I get some mess like 

travis

From this I cannot see neither what tests failed, nor why.

A somewhat better result would be from replacing all "\r" with "\n", however I still don't see much more useful information (in the img the test command is ran twice, with the bottom having endlines replaced).

travis-nl

Now if I instead look at the XML output I see the proper information --- what failed and why

travis-pig

The syntax is highlighted with [[pygmentize](http://pygments.org/), but raw XML output should be just a step, not the final result.

So from the XML we can later produce some much nicer reports, which would not be possible with the default output.

> Unfortunately, Travis doesn't have a built-in/free build artifacts upload mechanism.

Well the user can still push the build artifacts out of travis via ssh/s3/mail/whatever; since the test result file is now available the user is free to consume in whatever way he likes.

So the way forward I see would be to keep the XML file there and in addition to that extract the useful information out of the XML and print it to stdout in some user-friendly format.

fniephaus:
Oh, I almost forgot how bad a failing Pharo build on Travis looks like.
I will add simple printing for now, and then we can produce nice outputs later.

Here's a pretty cool example (uses syntactical Travis-specific sugar to enable folding):
image
https://travis-ci.org/kattrali/xcpretty-travis-formatter/jobs/52970340#L121

Using the XML will also allow us to unify the build result style across dialects which would be nice.

PR #44, opened 2016-01-27
split sources in platform packages

estebanlm:
I made several things here: 

- refactored specs so they are just definition containers and delegate to platform implementations when they have something to do. 
- implement platform variants for Pharo, Squeak and Gemstone and moved all significant operation there (so each platform can use their variations).
- added #packages category 

For Pharo is working and I *suppose* it will work without much change in Squeak (but I didn't find where to add the squeak specific file, heh). For Gemstone is just presented... almost nothing implemented (but it was like that before).


estebanlm:
sorry, I opened this by mistake over master and not dev

PR #45, opened 2016-01-27
split sources in platform packages

estebanlm:
I made several things here:

- refactored specs so they are just definition containers and delegate to platform implementations when they have something to do.
- implement platform variants for Pharo, Squeak and Gemstone and moved all significant operation there (so each platform can use their variations).
- added #packages category

For Pharo is working and I suppose it will work without much change in Squeak (but I didn't find where to add the squeak specific file, heh). For Gemstone is just presented... almost nothing implemented (but it was like that before).


fniephaus:
Thanks for the PR, @estebanlm.
Since this is a bigger change, I will need to load this into different images and test it.
Unfortunately, I won't have much time over the weekend, but I will follow up on this as soon as I can.

fniephaus:
I had a look at the things that you have changed and I like the way you separated platform-specific code!
Here are some minor notes:

- Using `self allSubclasses anyOne` to detect the `platformClass` expects users to load the correct extension package (e.g. SmalltalkCI-*Pharo*), but what if the user loads everything? I decided to use `self allSubclasses detect: [ :cl | cl isPlatformCompatible ]` instead.
- Since STON is a hard dependency, I moved the code back to `SmalltalkCI class >> initialize`, but it will still delegate the installation to the corresponding platform-specific class.
- I don't think `SmalltalkCI` should know how to load a `SCIMetacelloLoadSpec`, but `SCIMetacelloLoadSpec` should ask `SmalltalkCI` for the working directory.
- I've re-added `SCISqueakTranscript` which was missing I think.

Also, is there anyway, to run the `CommandLineTestRunner` *and* export a JUnit XML? SmalltalkCI expects a XML file to be present after a build and also prints that file to the Travis log.

PR #46, opened 2016-02-03
Fix the default config by using 'loading' and not 'loadSpecs'

zecke:
In 5c16ba63b092de2f0debca406644be426a678557 the loadSpecs got
renamed to loading but the default file generation has not been
updated.


fniephaus:
Thanks for fixing this, @zecke!

Issue #49, opened 2016-02-05
TestFailure: Block raised NotFound: Object is not in the collection


dalehenrich:
While working on porting SmalltalkCI to GemStone, I accidentally used the following `.smalltalk.ston`:

```ston
SmalltalkCISpec {
  #loading : [
    SCIMetacelloLoadSpec {
      #baseline : 'SmalltalkCITests',
      #directory : 'tests',
      #platforms : [ #squeak, #pharo, #gemstone ]
    }
  ],
  #testing : {
    #exclude : {
      #classes : [ #Object]
    }
  }
}
```

and it hit an [error that is present for all dialects](https://travis-ci.org/dalehenrich/smalltalkCI/jobs/107343587#L280). I've [fixed the bug for GemStone](https://github.com/dalehenrich/smalltalkCI/commit/61f1591d86835dbffce39ba73b771b5ef1153cb8) and will be including the GemStone fix as well as a test for the bug in the pull request addressing Issue #48.

dalehenrich:
I've added the fix to the `SmalltalkCI-Core.st file, but apparently that file is not used by Squeak ... the [Pharo builds are green](https://travis-ci.org/dalehenrich/smalltalkCI/builds/107356054)...

fniephaus:
Thanks for fixing this! Squeak builds fail, because I need to update the images first. SmalltalkCI and Metacello are preinstalled in these images to speed up builds

fniephaus:
Has this been fixed now?

dalehenrich:
Yes, [this code was in my recent PR](https://github.com/dalehenrich/smalltalkCI/commit/ef5b7d2a854c39d79d8f234b367df908330bb8de) ... Of course the test may not be running as the tests for this are in the [dead code](https://github.com/hpi-swa/smalltalkCI/issues/69)

Issue #48, opened 2016-02-04
SmalltalkCI>>readSTONSpec: is not portable to GemStone


dalehenrich:
The classes StandardFileStream and MultiByteFileStream used in [SmalltalkCI>>readSTONSpec:](https://github.com/dalehenrich/smalltalkCI/blob/master/src/SmalltalkCI-Core.st#L280) are not present in GemStone and the method probably should be made abstract and the current implementation moved to Squeak, as it appears that SmalltalkCI>>readSTONSpec: is implemented differently for Pharo as well ...

Now is a good time for me to look into this as well as fleshing out the rest of the GemSTone implementation.

BTW, I thought that I saw a FileTree repo for the SmalltalkCI code ... should I wait until that code is merged before making changes .... I haven't had to work with chunk format files since the 1980's:)

dalehenrich:
Okay, what I thought was a move of the smalltalkCI code into packages "package support" was not what I thought it was ... you realize that I will have to hand edit these chunk files to avoid massive changes when filed out from GemStone :(

dalehenrich:
I think I'll create a FileTree repo somewhere with this code in it and then let you create chunk files :)

dalehenrich:
Hmmmm, I can't even edit the .st files with a test editor because of line ending issues (doing development on a mac) ....

dalehenrich:
... I'll edit the chunk files in the web browser

fniephaus:
No problem at all, I'm happy to take care of the rest.
`.st` files aren't really nice to develop with, but they are easy to deploy since they don't require anything to be installed in an image.

dalehenrich:
Thanks, I'm just about ready to load/run the test packages, which I presume are only intended for testing the SmalltalkCI implementation (they don't appear to be run during the actual travis-ci run from the test scripts although, I still can't tell what's actually being tested there:). Shouldn't these tests be run by travis (I can't tell if they are or not...)

I assume that there are at least a handful of additional changes that I will need to be make for portability... ...

fniephaus:
Yes, the Smalltalk tests in `./tests` are self-tests and Travis loads and runs them when testing smalltalkCI itself. The code coverage isn't really high at the moment as the code is still changing a lot, but I hope this all will change soon.

dalehenrich:
[...]

PR #51, opened 2016-02-07
Make smalltalkCI compatible to GemStone

dalehenrich:
Fix for Issue #48 ... SmalltalkCI ported to GemStone:

- includes a [fix](https://github.com/dalehenrich/smalltalkCI/commit/ef5b7d2a854c39d79d8f234b367df908330bb8de) for Issue #49
- .st files converted from CR to LF 
- the GemStone SmalltalkCI code is stored in a filetree repository and there is no reason why Pharo and Squeak don't share the code here - after all the tests are all shared in filetree... I've made several bug fixes beyond Issue #49, but there is no practical way for me to share those fixes other than through sharing the code itself. @estebanlm, if you want to share code for SmalltalkCI vi filetree, we can find a better location for the code repository and share it ... 
- significant refactoring of the SmalltalkCI tests so that it's one test case per spec combo
- ported SCITestReport to GemStone to produce the proper xml result file 
- added #projects test type, that takes list of Metacello project names ... this code should be able to run in any dialect with Metacello installed 
- add a #'*' test type that causes all subclasses of TestCase to be added to the classes list. the #include and #exclude lists are used ... updated README
- add #onWarning to load spec, so that Warnings get logged to Transcript without having to add explicit handler
- added a number of GsDevKit/tODE artifacts in support of SmalltalkCI, including a [runSmalltalkCI script](https://github.com/dalehenrich/smalltalkCI/blob/issue_48/gemstone/gsdevkit/bin/runSmalltalkCI) that runs a SmalltalkCI spec against a given stone ...
- changed top-level .gitignore to exclude the common editor data files
- @estebanlm, [when a project is specified, the #packages entry is ignored](https://github.com/dalehenrich/smalltalkCI/blob/issue_48/tests/SmalltalkCI-Tests.package/SmalltalkCITest.class/instance/testPharoClassesTestPackages2.st) in Pharo, but I think that the #packages and new #projects entry should behave the same as #classes and #categories, in that they add a list of tests to the list of classes in addition to the classes in the loaded project. I've added a [test to illustrate differences between GemStone and Pharo](https://github.com/dalehenrich/smalltalkCI/blob/issue_48/tests/SmalltalkCI-Tests.package/SmalltalkCITest.class/instance/testGemStonePharoClassesTestPackages1.st))


dalehenrich:
The test: SmalltalkCITest>>testGemStonePharoClassesTestPackages2 is failing in GemStone currently failing (just discovered while creating SmalltalkCITest>>testGemStonePharoClassesTestPackages1)  so I will need to supply a fix before this PR is merged ...

dalehenrich:
Looks like [Pharo fails the test as well](https://travis-ci.org/dalehenrich/smalltalkCI/jobs/107608938#L292) ... so it appears to be a test bug unless `SmalltalkCI.*` is supposed to be a valid package pattern

dalehenrich:
Pharo tests should be passing again ... squeak tests await your reloading the .st files into the standard squeak image ... and the GemStone tests await integration of GemSTone into container-based builds ... SmalltalkCI independently tested as [I'm currently using a jury-rigged version of GemStone to test tODE](https://travis-ci.org/dalehenrich/tode#L2544) ... also an example of a project with 1000's of tests and the utility of a [failure summary](https://travis-ci.org/dalehenrich/tode/builds/107594163#L3957) at end of log

dalehenrich:
The [pharo test for testGemStonePharoClassesTestPackages2 is failing](https://travis-ci.org/dalehenrich/smalltalkCI/jobs/107610773#L115), not sure why ... it passes for gemstone (in my in-image tests)

dalehenrich:
@estebanlm, here's the skinny:

```
 SmalltalkCI-* fails for Pharo, passes for GemStone
SmalltalkCI.* fails for GemStone, passes for Pharo
```

I assume that both patterns should pass for both platforms unless there is a good reason that the `SmalltalkCI-*` pattern should fail on both platforms ...

estebanlm:
On 08 Feb 2016, at 00:02, Dale Henrichs <notifications@github.com> wrote:
> 
> @estebanlm <https://github.com/estebanlm>, here's the skinny:
> 
>  SmalltalkCI-* fails for Pharo, passes for GemStone
> SmalltalkCI.* fails for GemStone, passes for Pharo
> I assume that both patterns should pass for both platforms unless there is a good reason that the SmalltalkCI-* pattern should fail on both platforms ...
> 
> 

I don’t know… “*" should pass in both platforms (while “.*” is regex and then just supported, I think, by pharo).
I will take a look this week.

Esteban

dalehenrich:
@estebanlm I'm going to put in a conditional expected failure for these tests and open a separate issue ... once you've debugged the pharo side, I'll follow your lead and take a similar approach for GemStone ...

I did a few experiments with Pharo and and both forms are apparently `legal regex`, `SmalltalkCI-*` doesn't match any packages when run as regex ...

fniephaus:
The `smalltalk.rb` is now in production, but unfortunately, it still doesn't work:
https://travis-ci.org/hpi-swa/smalltalkCI/jobs/107904298#L1227

@dalehenrich do you have any idea what needs to be done to get this working?

dalehenrich:
The root cause is this error:

```
License has cpuAffinity restriction,
 sched_setaffinity mask 3 failed with errno 22, Invalid argument
'
STDERR: 
startstone[Error]: Stone process (id=7071) has died.
startstone[Error]: Examine '/home/travis/build/hpi-swa/smalltalkCI/GsDevKit_home/server/stones/travis/logs/travis.log' for more information.  Excerpt follows:
```

and I can make changes in the run.sh to work around this problem ... (we had a similar issue using builderCI so I know what needs to be done) ....

fniephaus:
@dalehenrich There currently is an issue with GCE builds on Travis (#59), so we are already working on the next `smalltalk.rb` update.
I understand that everything that requires `sudo` is already part of the `smalltalk.rb` and everything else can now be done without having to bug the folks at Travis? Is that correct?

dalehenrich:
Resolved the key file problem, but we're getting the same failure as before (I think):

```
Loading and testing project...
0.35s[Info]: libssl-3.2.12-32.so: loaded
Error: Unable to create a GemStone session, check netldi log file.
NetLDI service 'travis_ldi' not found on node 'testing-worker-linux-docker-eecf9117-3369-linux-1.prod.travis-ci' port 60519 :
getaddrinfo failed, EAI error -2, Name or service not known,  For further information about login failures, check the gem log file
```

Here's the tale of the host name tape:

```
Are the host names set up correctly for GemStone?
testing-worker-linux-docker-eecf9117-3369-linux-1
172.17.7.163	testing-worker-linux-docker-eecf9117-3369-linux-1.prod.travis-ci.org testing-worker-linux-docker-eecf9117-3369-linux-1
127.0.0.1	localhost testing-worker-linux-docker-eecf9117-3369-linux-1
::1	localhost ip6-localhost ip6-loopback testing-worker-linux-docker-eecf9117-3369-linux-1
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
Well?
```

Now the host name that the stone is trying to connect to is: `testing-worker-linux-docker-eecf9117-3369-linux-1.prod.travis-ci` which is odd, since that name doesn't show up anywhere in the hosts file or as hostname ... is there yet another way to define the host name?

Perhaps the description file can be jiggered?

fniephaus:
How about you get rid of `.prod.travis-ci` at the end?

dalehenrich:
Well I have to track down the source of this host name in the 
GsDevKit_home stack ... for some reason, "hostname" is not being used to 
acquire the name of the host ... good news though is that the source of 
this is in code that is independent of the GemStone release ... so I 
don't think we'll need any additional `sudo` repairs :)

Dale

dalehenrich:
@fniephaus what are rules for exit status processing in the run.sh scripts?

I'm seeing scripts fail in the gemstone run.sh script but the [run.sh script itself does not exit with an error](https://travis-ci.org/dalehenrich/smalltalkCI/jobs/110228094#L1430) ... it just keeps on running which is of course is a total waste ... despite the `set -e` at the top:

I want the run.sh script to fail if any script running fails .. but I see in the squeak run.sh script that you are manually checking return status ...

fniephaus:
@dalehenrich there's no rules really. If you don't want smalltalkCI to continue, make sure to return something != 0 here:
https://github.com/hpi-swa/smalltalkCI/blob/ca8bc5c1183df6946480129866bcd67810c0a08d/gemstone/run.sh#L102

Fail as early as you want. Make `prepare_stone` check and return an error and then check the error in `run`:
```bash
...
gemstone::prepare_stone "${stone_name}" "${config_smalltalk}" || exit_status=$?
gemstone::load_and_test_project "${stone_name}" || exit_status=$?
...
```

Also, there now is a `repository` direction:
https://github.com/hpi-swa/smalltalkCI/tree/ca8bc5c1183df6946480129866bcd67810c0a08d/repository
Could you pull the code from master and merge your GemStone code with this?

dalehenrich:
So the script itself does not exit with a non-zero exit status? I'm confused why the `set -e` does not work like it does in a regular bash script ... 

I did see that you'd gone with a filetree repo, so I'll do the merge ... 

Still tracking down issues that were masked by the script not exiting when intermediate scripts were failing ... lost time thinking earlier steps were completing without errors ...

fniephaus:
Sorry about the confusion :(
The platform-specific `run.sh` scripts are being [sourced](https://github.com/hpi-swa/smalltalkCI/blob/99e03d360a50dc65c45290d9a77a1b1c699275c1/run.sh#L331) into the main `run.sh` and then `run_build()` is invoked. And the current `run.sh` for GemStone does not exit at all.
Hope this makes sense. Let me know if there's something I can help with.

dalehenrich:
I still don't understand why `set -e` does not have an effect ... it's a pain to have to add:

```
if [[ ! $? -eq 0 ]]; then
        print_error_and_exit "blah blah"
    fi
```

after every shell command in the `run;sh` script, but that appears to be necessary?

fniephaus:
> `set -e` When this option is on, if a simple command fails for any of the reasons listed in Consequences of Shell Errors or returns an exit status value >0, and is not part of the compound list following a while, until, or if keyword, and is not a part of an AND or OR list, and is not a pipeline preceded by the ! reserved word, then the shell shall immediately exit.

You could also just do a `... || exit 1` at the end of each line.
But yes, this can in fact be a problem and I will need to make it better.

dalehenrich:
[**Major milestone** - got to point where tests are run and some of them are failing :) ](https://travis-ci.org/dalehenrich/smalltalkCI/jobs/111127587#L2689)

fniephaus:
This is awesome! This still counts as a green build I'd say, because none of the SmalltalkCI tests are failing :smile: I actually enabled the `STON-Tests` yesterday -.-
Anyway, I'm confident that we can get this in this week. Yay!

dalehenrich:
The main task is merging my SmalltalkCI packages (formerly .st files) with yours ... and getting the STON tests to pass ... but we're out of the woods for sure!

fniephaus:
@dalehenrich I will try to merge this tonight if you are not already working on it.

fniephaus:
Also, I believe you have ported STON to GemStone, right? Do we need to use that then?

dalehenrich:
@fniephaus I'll do the merge ... it is likely that Squeak and GemStone can share the SmalltalkCI-Report code (that's where I added support for expected failures) but I will leave that exercise to you ... but I will move my SmalltalkCI-Report package and you can determine which parts are sharable ...

fniephaus:
perfect! Let me know when I can merge this :)

dalehenrich:
Yes, STON is already present in GemStone and I'll take care of that in the BaselineOfSmalltalkCI (STON only needs to beloved for Pharo and Squeak) ...

dalehenrich:
@fniephaus what are the sudo rules for OSX? I may be experiencing failures on OSX due to the shared memory not being set correctly on OSX ... are you currently doing any shared memory setup for osx in your smalltalk.rb script? ... The [shared memory setup for both linux and osx is here](https://github.com/GsDevKit/GsDevKit_home/blob/master/bin/utils/installOsPrereqs#L145-L300)

fniephaus:
Shared memory should be set up:
https://travis-ci.org/dalehenrich/smalltalkCI/jobs/111362797#L17
[This](https://github.com/travis-ci/travis-build/blob/38f884a97a966291a37bbf44d79904f48215d452/lib/travis/build/script/smalltalk.rb#L160-L188) is what happens.
However, there're two `HostName: not set` errors. Could those be responsible for failing OSX builds?

dalehenrich:
Yeah, I think the hostname is implicated:

```
Stone was unable to attach to shared cache on localhost
Reason: fork of cache monitor failed, Monitor process (977) did not start.
; NetLDI service ''gs64ldi'' not found on node ''localhost6'' port 50377 :
connect(10,::ffff:127.0.0.1,port=50377) failed with errno=22,EINVAL, Invalid argument (programmer error)
```

I'll try to do a little debugging of the problem and try to identify specifically what might be wrong .... of course I took the logging out of the run.sh script, so I've got to put it back ... the OS X builds are not a high priority right now as I want to get my big set of changes integrated first ...

dalehenrich:
Once I characterize the mac issue. I'll open a separate issue so we can merge in my changes ...

fniephaus:
That's fine!

dalehenrich:
Okay the OS X hostnames are not set correctly:

```
Are the host names set up correctly for GemStone?
Traviss-Mac-441.local
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1	localhost 
255.255.255.255	broadcasthost
::1             localhost  
fe80::1%lo0	localhost
Well?
clone_gsdevkit
```

The result of `hostname` is `Traviss-Mac-441.local` and `Traviss-Mac-441.local` does not show up in /etc/hosts....

fniephaus:
Ok, I can get that fixed. There's no way GemStone could just use `localhost`?

dalehenrich:
we call `hostname` to find out the name of the host ... the code that is involved creates a file that uses the hostname to access a remote host, so wiring in `localhost` is not really an option - travis-ci is not the primary application here :)

dalehenrich:
By jove, I think we've got it!

dalehenrich:
Oops, [celebrated too soon](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/111388569#L2995) ... looks like Issue #65 can ~~fail~~pass for the wrong reason ... I moved the GemStone directory into `_build` and if your intent is to run a different set of tests it is overkill to delete the whole build ... in this particular case I think that I can fix the problem that caused the ~~failure~~pass, but as I said, it seems to be expensive to delete the whole thing ... there is a less expensive alternative that could be used for GemStone -- restore from a backup taken before loading the SmalltalkCI tests ...

dalehenrich:
on second thought rebuilding for Issue #65 is not so bad, however a false positive is not good ...

dalehenrich:
It also looks like my workaround for Issue #65 is working, so I imagine that we can go ahead an pull trigger on the merge!

dalehenrich:
ah, my good friend API rate limit ... I will probably go after the API rate limit problem  with a vengeance now :) ... not able to work on this most of today though...

fniephaus:
I will just restart them later then ;)

fniephaus:
Are you online somewhere, @dalehenrich?

Issue #69, opened 2016-02-25
SmalltalkCI-Tests remains in ./tests


fniephaus:
What do we do with these tests?
https://github.com/hpi-swa/smalltalkCI/tree/ee4070f14b153fd85d6ad7f65a150a18efd7d4ae/tests/SmalltalkCI-Tests.package/SmalltalkCITest.class/instance

dalehenrich:
Ah ... these tests are supposed to be run ... and they were passing for all three platforms in my checkout  .... I see that they are not being run now so I will look into what is happening ...

dalehenrich:
Aha ... the tests repo got move from `.tests` to `.repository` and I didn't notice --- no conflicts --- and my new tests were not merged .... I'll take care of that soonish

dalehenrich:
action has moved to PR #74

PR #74, opened 2016-02-26
Fix for Issue #69

dalehenrich:
The new tests that I wrote were mainly aimed at testing the package-related test creation features which are only for GemStone or Pharo ... presumably it may be useful for Squeak to support pacdkage-based specs as well ... if not the gemstone/pharo tests could be split into separate packages that are only loaded for gemstone/pharo ....


fniephaus:
+1 for platform-specific test packages. That'd make a lot of sense, because we also need to test platform-specific code.
I will merge this in now and we can add more tests in a next step.

Are any of your tests NOPs on Squeak or do they happen to work on Squeak as well?

dalehenrich:
The tests with gemstone and/or pharo in the name are the platform-specific tests and are noops (passes) on Squeak ...

fniephaus:
Ok, so we should move them to platform-specific packages.

dalehenrich:
I can do the work ---- do you want ot open a separate issue?

fniephaus:
Another PR would be enough, I think. Thanks! :)

dalehenrich:
Haha, but I need a reminder somewhere ... i=unless you want to keep asking me when I'm going to do it:)

Issue #64, opened 2016-02-22
break the build if Undeclared detected


DamienCassou:
By default, Monticello reports a warning if it is loading a method that references a class not in the system. The warning is non blocking which allows dependencies to be loaded after the package depending on them. The problem is that this process allows packages to be badly defined with non-declared or circular dependencies. Would it be possible to have an option to transform these warnings into build errors?

fniephaus:
Interesting idea. I guess we could add support for something like `#failOnWarnings` to `SCIMetacelloLoadSpec` and then you could enable it by setting the value to `True`.

I wonder what @dalehenrich thinks about this.

BTW: I literally just made smalltalkCI ignore warnings (0a1f772031675d245a87acf8778730ad57e9d648) :grin: .

dalehenrich:
My thoughts are that we'd like to fail on specific warnings rather than generic warnings ... @DamienCassou could you supply a stack that shows an example of the Warning you want ignored? If this is coming from Metacello, I'd be willing to create a specific class so that we can selectively fail ...

DamienCassou:
I have the impression the warning comes from the compiler, in particular the class Encoder which signals an UndeclaredVariableWarning. This exception can be resumed. In Pharo, when Monticello loads a package containing a method with a reference to a non existing class, a message is shown in the transcript and loading continues. In my example, I have a package Foo-Main with a class FooMain with a method foo referencing a class in another package. If this dependency is not loaded, loading Foo-Main results in the following being shown in the transcript:

    FooMain>>foo (FooDependency is Undeclared)

DamienCassou:
Metacello, in `MetacelloMCVersionValidator>>validateCleanLoadAndTestsForVersion:loads:` has code related to this exception as well.

dalehenrich:
As I think about the UndeclaredVariableWarning, I'm pretty sure that the Warning can be thrown in cases where the global _will_ be resolved by another package being loaded a bit later during the same project load ... so it is not fool proof ...

I deal with this in GemStone by having a test that detects `sent but no implemented methods` and `undeclared globals` ...

DamienCassou:
Marcus told me the class Encoder is not used anymore, it's the old compiler. What is important now is that the new compiler signals `OCUndeclaredVariableWarning` when a method references an undeclared variable. I would like the code loader to signal an error if this warning is signaled. Is that possible?

DamienCassou:
[...]

PR #79, opened 2016-03-05
Force loading of packages during merge warnings

theseion:
Fuel is already loaded in Pharo. During load of a newer version Monticello issues an `MCMergeOrLoadWarning` (only exists in Pharo). SmalltalkCI doesn't correctly handle the warning, which results in a mixture of old and new code being loaded.

This fix forces a clean load of the new version, discarding changes in the loaded package.


fniephaus:
it seems that you are only changing meta information? Or am I missing something?

theseion:
Yeah... Forgot to add the change :)

repository/SmalltalkCI-Pharo.package/SmalltalkCIPharo.class/instance/loadProjects.st
```diff
@@ -0,0 +1,6 @@
+as yet unclassified
```
fniephaus: Would you be so kind and add this to a category?

fniephaus:
Are you sure that this only needs to be done in Pharo?
@dalehenrich for a second opinion?

dalehenrich:
The "merge or load warning" is Pharo only ... 

The #onWarningLog attribute could be used to allow the load to proceed instead of hard-wiring an exception handler (this is no longer GemStone-specific after my  recent merge --- ought to update README). 

A handler for just the "load or merge Warning" is pretty  specific and won't mask other Warnings that might be more interesting ... Perhaps it would be useful to log the fact that the exception was handled before resuming?

There's an outside chance that it might be desirable to break the build as discussed in Issue #64.

For the long term perhaps we could add a list of Warnings to be handled and resumed in addition to the [fail load on list](https://github.com/hpi-swa/smalltalkCI/issues/64#issuecomment-188481355) and deprecate the #onWarningLog attribute.

repository/SmalltalkCI-Pharo.package/SmalltalkCIPharo.class/instance/loadProjects.st
theseion: Will do.

theseion:
> Are you sure that this only needs to be done in Pharo?

For Fuel yes. In general I expect that MC will issue a `Warning` instead of the more specific `MCMergeOrLoadWarning`.

> A handler for just the "load or merge Warning" is pretty specific and won't mask other Warnings that might be more interesting

True. However, `MCMergeOrLoadWarning` expects `true` when resuming, which is not the default resumption value. So this warning needs special handling.

> Perhaps it would be useful to log the fact that the exception was handled before resuming?

Yes. I had to guess my way to fixing the problem because nothing was visible in the log. It would be nice to see warnings and exceptions, even if they are then being handled.

> There's an outside chance that it might be desirable to break the build as discussed in Issue #64.

I think such a setting is be a good idea (and maybe should be the default). However, for Fuel specifically I must be able to force the load, and there's no way to disable the warnings (unless I patch MC in a `#preLoadDoIt`...)

> For the long term perhaps we could add a list of Warnings to be handled and resumed in addition to the fail load on list and deprecate the #onWarningLog attribute.

While that sounds like a good idea, as I state above I fear that MC in Squeak will issue a `Warning`. So handling of that specific case would entail the same handling for all warnings.
In addition, we would need an option to specify the handling behaviour, as `MCMergeOrLoadWarning` expects a specific value when resumed.

theseion:
I can confirm that Squeak uses `Warning` in `MCVersionLoader>>warnAboutLosingChangesTo:` via `Object>>notify:`

fniephaus:
I've merged this now, so that @theseion can test Fuel.
Could one of you please update #64 or open a new issue to follow up on the MC warning discussion?

theseion:
Thanks. I've opened issue #80.

Issue #83, opened 2016-03-17
Cache bootstrapped version of GemStone by GemStone version number


dalehenrich:
The bootstrapped version of GemStone changes relatively infrequently and if the `--clean` option does what I think it does, a developer can clean their cache whenever they are interested in picking up the latest bootstrapped build ...

fniephaus:
Where does the bootstrapped version of GemStone live? `_builds/GsDevKit_home`?
It is very easy to [cache a directory](https://docs.travis-ci.com/user/caching/#Arbitrary-directories). Maybe try caching `$SMALLTALK_CI_BUILD/GsDevKit_home`?

fniephaus:
> There is one cache per branch and language version/ compiler version/ JDK version/ Gemfile location/ etc.

See https://docs.travis-ci.com/user/caching/#Fetching-and-storing-caches

dalehenrich:
Right now it seems to make the most sense to simply cache the extent file produced by the initial createStone ... Also the gsDevKitCommandLine pharo image/changes files are also worth caching ... Finally, since the pharo vm, image and chages files are being cached, I will cache those as well (if they haven't already been cached) ....

dalehenrich:
So smalltalkCI itself is not using the `_cache` directory as a dependency cache?

fniephaus:
Nope, not yet. It is fairly easy to enable this, but I haven't had the time to check how caching could speed up GemStone builds. Apparently, you were able to speed it up quite a bit!
For Squeak and Pharo, I don't think that caching would help much, because almost everything is just downloaded and that's it. Directory caching works the same way (Downloads everything and then extracts it).

dalehenrich:
Yeah, if the travis timings are to be believed, [the non-cached timings](https://travis-ci.org/dalehenrich/smalltalkCI/builds/116756215) vs [cached timings](https://travis-ci.org/dalehenrich/smalltalkCI/builds/116736746) are actually faster (up to 2x slower for  some Squeak and Pharo builds).

I'm guessing that those differences are showing up because changes were found in the cache for those builds ... not sure why, but I was only concerned with whether or not the non-GemStone builds were passing not why they were running slower ... the cache upload time is pretty slow ...

fniephaus:
I've enable caching for GemStone builds on Linux, but oddly enough the cache size is 0?
[...]

PR #87, opened 2016-03-17
Fix for Issue #83

dalehenrich:
Per #83, added code to gemstone/run.sh to cache extents and gsDevKitCommandLine image (and Pharo3.0 download) when dependency cache is enabled in `.travis.yml`:

```yml
cache:
  directories:
    - $TRAVIS_BUILD_DIR/_cache
```
When running local script the env var `GS_TRAVIS_CACHE_ENABLED` can be used to control caching behavior - the default is to use the cache, so the following will turn off the cache:

```
export GS_TRAVIS_CACHE_ENABLED=false
```

The above code has been included in the `.travis.yml` file for smalltalkCI (commented out), if dependency caching is enabled gemstone uses 126MB zipped (includes a Pharo-3.0 image. changes and vm). The squeak and pharo images and vms add another 200MB (314MB cache size according to travis for full build), but I think there's some cruft in there as the squeak builds leave tgz files for images and vm in the `_cache` directory. If that is not an acceptable cache size then we may want to leave the code commented out (or removed) ... but if the cache isn't too large the GemStone linux builds are about 3 times faster ....

I am leaving  the expected failures for the GemStone OSX builds enabled. The api rate limit is no longer an issue, but I am still seeing occasional network errors causing build failures on the OSX builds for GemStone,restarting the build can get past those issues, unlike the api rate limit which seemd to be perpetually exceeded on the OSX boxes -- but it is annoying to get these random failures.


dalehenrich:
I've got to head out, but it looks like there might be a logic error in the gemstone/run.sh script ... I will try to check later tonight ...

fniephaus:
Thanks, I will try to review this tomorrow.

dalehenrich:
Yep, I had a typo that caused env vars controlling caching to be set incorrectly - `gemstone/run.sh` was updating the `_cache` dir without dependency caching actually being enabled - fixed typo and verified the logic was correct ...

dalehenrich:
[build failure](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/116809819#L1545) is a network error --- restart and it should pass

gemstone/run.sh
```diff
@@ -109,6 +179,18 @@ run_build() {
       ;;
   esac
 
+  # To bypass cached behavior for local build, export GS_TRAVIS_CACHE_ENABLED
+  #   before calling run.sh
+  if [ "${GS_TRAVIS_CACHE_ENABLED}x" = "x" ]; then
```
fniephaus: Why is this necessary? One can easily delete the cache by removing the `_cache` directory.
I implemented caching for Squeak and Pharo, because I didn't want to download everything for every build. So that is very useful when I run builds locally to debug a problem.

dalehenrich: My main thought was that I wanted to be able to test the code locally when `GS_TRAVIS_CACHE_ENABLED=false`, since that is a separate code path in gemstone/run.sh.

fniephaus: I'm happy to leave it in if it helps :)

Issue #75, opened 2016-02-26
Add platform-specific test packages


fniephaus:
Since we have platform-specific packages for smalltalkCI, we also need platform-specific test packages.
@dalehenrich has already started adding platform-specific test, that we should move first.

dalehenrich:
any additional work should be done under the cover of a new issue

PR #90, opened 2016-03-18
initial cut at Issue 75

dalehenrich:
This is the initial cut at multi-platform tests for Issue #75. There are still a few places that have code based upon platform attributes, but I thought that the bulk of the test was common to all platforms and that breaking out the platform bits would just create a much more complicated package structure ... future refactorings could go further in this area ...


fniephaus:
Can this be merged already or do you still work on this?

dalehenrich:
The tests are all passing ... and there _are_ additional tests, so let's go ahead and merge this ... I think it could be refactored a bit more at the cost of additional packages and more convoluted code in the tests ... or a more complete revamp :)

fniephaus:
Cool, thanks!

Issue #52, opened 2016-02-07
Customize the `reproduce` message for GemStone


dalehenrich:
[This message](https://travis-ci.org/dalehenrich/smalltalkCI/jobs/107610773#L135) would *work* for GemStone, but since run.sh does a full install of GsDevKit_home it is overkill for someone who may already have a GsDevKit_home installation ...

fniephaus:
If we make sure that smalltalkCI does nothing if GemStone is already installed, then there is nothing to do I guess.
The message is automatically generated: https://github.com/hpi-swa/smalltalkCI/blob/add3c7fd5f64473fdd45fe921c5b30ad1d6da3af/helpers.sh#L66-L75

dalehenrich:
That's a very good point ... duh!

dalehenrich:
please assign this to me and I will work on that ...

fniephaus:
This is now the case:
https://travis-ci.org/hpi-swa/smalltalkCI/jobs/112107806#L4022

However, I don't really feel comfortable showing steps to reproduce for a GemStone build, because it requires some OS prerequisites and it does way more than just downloading a vm and an image.
What do you think, @dalehenrich?

dalehenrich:
@fniephaus it might turn out that the "solution" might be custom GemStone options passed in at the ./run.sh command ... isn't their a "posix style" for specifying options to an underlying command? Such an *escape clause* would make it possible to simply add custom options to the gemstone/run.sh command, allowing for the re-use of existing checkout or ???

fniephaus:
I think for now, the good ol' Smalltalk approach should be sufficient: namespace all the GemStone options with ```--gs-``` (e.g. ```--gs-custom-option```). :smile: What do you think?

dalehenrich:
@fniephaus ... but then that would require that we keep run.sh options and gemstone/run.sh options in synch ... rather than a pass-through option ...

fniephaus:
The gemstone/run.sh is being sourced rather than called in a subshell. That means, smalltalkCI just needs to "accept" --gs-* (it fails if it doesn't know an option) and then there could be a second round of argument passing in the gemstone/run.sh. This way, nothing needs to be kept in sync and you are free to add custom options for GemStone. Sounds good? If yes, I could start working on this next week :)

dalehenrich:
[...]

PR #95, opened 2016-03-23
Added steps to run SmalltalkCI locally [ci skip] [#refs 92]

peteruhnak:


README.md
```diff
@@ -172,6 +172,19 @@ OPTIONS:
...
+
+1. clone your repository
+  * a) from github: `git clone https://github.com/your/project.git`, or
+  * b) clone your local repo: `git clone /wherever/is/your/project`, or
+  * c) use your local repo directly
+2. cd into the cloned directory `cd project`
+3. download smalltalkCI: `wget -O smalltalkCI.zip https://github.com/hpi-swa/smalltalkCI/archive/master.zip`
+4. unzip it: `unzip -o smalltalkCI.zip`
+5. run smalltalkCI, provide dialect you want to test and your local config: `smalltalkCI-master/run.sh --headfull -s "Pharo-5.0" .smalltalk.ston`
+
```
dalehenrich: hmmm, following these instructions it seems that I end up with a smalltalkCI checkout in the top-level of my `project`, which I don't think is desirable ... 

I think it might also be worth mentioning that each `run.sh`execution  is accumulated in the _build directory...

peteruhnak: Because that's exactly how Travis behaves (and when you are trying to reproduce a problem it's best to reduce variables).

But I could mention that you can download/unpack the zip anywhere you want as long as you provide correct path to the `.smalltalk.ston`

dalehenrich: I understand, but part of the reason for documenting this independently (in my opinion) is that smalltalkCI can be used as CI for local projects (obviating the need to download Jenkins to run automated tests) and in this scenario you might want an independent git checkout of smalltalkCI so that you smalltalkCI setup can be used to run tests against a variety of projects...

in the end I think it's worth covering both scenarios ... 

Perhaps a link to the reproduce travis-problem scenario can be included in the ` to reproduce bug` message ... There will be additional options for launching builds for GemStone (see #52) and this will turn out to be a tidy spot for documenting those options when they apear

peteruhnak: >
> in this scenario you might want an independent git checkout of smalltalkCI so that you smalltalkCI setup can be used to run tests against a variety of projects...
>
>
You mean git clone of smalltalkCI instead of downloading the zip?

dalehenrich: Yes, if a developer starts with a git clone of smalltalkCI, it is much easier to keep up with the changes to the project **and** they are more likely to actually contribute to the project:)

If they are doing a one-off reproduce bug in travis, the download makes sense, but if they are going to use smalltalkCI for CI, then a git checkout makes more sense ... in my opinion :)

dalehenrich: Does it make sense to mention that one should remove the smalltalkCI after you're done (if you've downloaded it into your project directory)

peteruhnak: Why? So you don't accidentally commit it? Are people not looking at what
they are commiting?

dalehenrich: No, because you are instructing folks to drop a big fat directory into their checkout and you should finish the job by having a cleanup step ... 

If it were up to me I would not drop the directory there --- there is really nothing to be gained by dropping smalltalkCI into a persons checkout ... smalltalkCI works fine from anywhere --- the project under test directory is calculated relative to the .smalltalk.ston file

If you have folks install smalltalkCI in a neutral location it can be used for multiple projects and/or over a long period of time and for multiple travis error conditions over time ... dropping smalltalkCI into a checkout means that it will have to be deleted before the next commit of their project something that will happen as soon as they solve their original issue ...

fniephaus: Let's not overthink this :)
It's really up to the user if he wants to download or git clone smalltalkCI, so why not mention both?
Once on a local filesystem, the user can initiate a build via `/path/to/smalltalkci/run.sh -s {SMALLTALK_VERSION} /path/to/my/projects/smalltalk.ston` (needs double-checking; I'm on a road trip, so I'm writing this from the top of my head).
Running Pharo and Squeak builds should be safe, I wouldn't recommend running GemStone except if you know what you are doing ;) (GemStone has more OS requirements).

fniephaus: @dalehenrich was faster -.-
And again, it's really up to the user where he wants to safe it. It can just live in the downloads or a tmp folder, but it might be even handy to have it in a convinient directory which is easy to reach, or even better: have a Bash alias point to the run.sh :D (maybe we even need an installer that does that and keeps smalltalkCI up to date :D)

dalehenrich: Regarding "not recommending running GemStone", the standard install script installs the OS prereqs (using `sudo`) so if they allow sudo access GemStone will function correctly ... 

With that said, I fully intend to extend whatever documentation gets written with the "customization for GemStone" (issue #52), whenever that gets figured out :)

fniephaus:
How do you like 76dd4e2fef7012d33bb79d1a5e1ac3cdc51b601d?

README.md
fniephaus: *I* can't "recommend" it, because I have never tried to run it locally or on any Linux or OS X machine, but that's not the point. I guess I just didn't like that a user is not being informed that or why `sudo` is used, but we are working on that :)

Issue #102, opened 2016-03-30
cannot interrupt clean prompt? ... if this is standard bash behavior it's acceptable


dalehenrich:
Here's what happens when I hit `CTL-c` at the prompt during clean:

```
foos:smalltalkCI>./run.sh --clean
builds at '/export/foos1/users/dhenrich/git/smalltalkCI/_cache'.
cache at '/export/foos1/users/dhenrich/git/smalltalkCI/_builds'.
Are you sure you want to clear builds and cache? (y/N): ^CsmalltalkCI has been interrupted. Exiting...

Continue with build progress? (y/N): ^CsmalltalkCI has been interrupted. Exiting...
^CsmalltalkCI has been interrupted. Exiting...
^CsmalltalkCI has been interrupted. Exiting...

foos:smalltalkCI>
```

I had to hit a return after `CTL-c` and even then the `CTL-c` did not appear to work ...

Also note that the directories are reversed in the `--clean` prompt:

```
builds at '/export/foos1/users/dhenrich/git/smalltalkCI/_cache'.
cache at '/export/foos1/users/dhenrich/git/smalltalkCI/_builds'.
```

should be:

```
cache at '/export/foos1/users/dhenrich/git/smalltalkCI/_cache'.
builds at '/export/foos1/users/dhenrich/git/smalltalkCI/_builds'.
```
 ... I've got a fix for this in my upcoming PR ...

dalehenrich:
I've fixed the part of this bug that I can fix .. the CTL-C is beyond my bash chops to address, so it's time to close or assign to someone else :)

fniephaus:
Can't you (re)assign issues as a contributor?

dalehenrich:
That must be a new feature as well ... the last time I checked (weeks ago) I couldn't change the assignee, nor could I close issues that I did not create ... and it appears that I can do that now as well ...

fniephaus:
That was why I added you as a contributor in the first place ;) just trying to make our lives as easy as possible

dalehenrich:
Haha, that's great .... now that I know I have different super powers I will try to look for new opportunities to use them :)

BTW, I am thinking that I don't want to overuse the privilege of being able to create a merge branches on the hpi-swa repository ... If I do work in my fork, you don't have to ignore it ... however, if I do work on a branch in hpi-swa, then you can easily contribute changes or test out intermediate results without too much trouble ... so what are your guidelines?

PR #103, opened 2016-03-30
Infrastructure for GemStone-specific eun.sh args

dalehenrich:
- Remove the --gs-* args that don't apply quite yet
- Partial fix for Issue #102 - fix typos for the `--clean` messages
- changes needed to make the `--gs-*` args code fuctional ...
- fix a GemStone-only bug in `--clear`

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -243,8 +243,8 @@ check_clean_up() {
   local question1="Are you sure you want to clear builds and cache? (y/N): "
   local question2="Continue with build progress? (y/N): "
   if [[ "${config_clean}" = "true" ]]; then
-    print_info "builds at '${SMALLTALK_CI_CACHE}'."
-    print_info "cache at '${SMALLTALK_CI_BUILD_BASE}'."
+    print_info "cache at '${SMALLTALK_CI_CACHE}'."
+    print_info "builds at '${SMALLTALK_CI_BUILD_BASE}'."
     read -p "${question1}" user_input
     if [[ "${user_input}" = "y" ]]; then
       clean_up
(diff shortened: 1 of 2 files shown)


fniephaus:
I will need to look into this tomorrow again. It's already pretty late here :)

dalehenrich:
no hurry ...

run.sh
```diff
@@ -270,6 +272,7 @@ clean_up() {
     fi
     if is_dir "${SMALLTALK_CI_BUILD_BASE}"; then
       print_info "  ${SMALLTALK_CI_BUILD_BASE}"
+      chmod -fR +w "${SMALLTALK_CI_BUILD_BASE}"
```
fniephaus: Why is this necessary? You are not doing the same with `${SMALLTALK_CI_CACHE}`.

fniephaus:
Could you check out 1667ebbbebcd21b1a5dccf817dd1742a3d6d8a22 and verify that `--gs-*` arguments are still working? I quickly added two dummy options and they are working for me:

image

run.sh
dalehenrich: The GemStone download is chock full of read only files and the `chmod` is required before `rm`ing the `_build` directory. There are no read only GemStone artifacts in the cache directory so it isn't required ...

dalehenrich:
If they work for you then I'm fine ... I'll test your changes after I merge master again ... I will  be adding some real options shortly, so I'll test for myself at that point in time ...

run.sh
fniephaus: Cool, I didn't know. Adding a comment...

PR #140, opened 2016-05-09
adjust list of supported GemStone versions ... should be no limits on which versions are supported

dalehenrich:


fniephaus:
I didn't know we support this many GemStone versions, great! :)

Issue #152, opened 2016-06-08
Add spec for loading MC packages


Uko:
I have use cases where it would be nice to specify a single package (or two…) in .ston spec. Maybe there can be an `SCIMonticelloLoadSpec` that can load single packages from repositories

fniephaus:
I assume you want to load packages without Metacello?
It should be fairly easy to add support for more loading formats such as a `SCIMonticelloLoadSpec` or even a `SCIFileTreeLoadSpec`...

Uko:
Would it make sense to have something like

SCIGoferLoadSpec {
      #repository : 'http://…',
      #packages : [ 'Package1', 'Package2' ]
    }
 
that will translate directly into Gofer commands?

fniephaus:
A `SCIGoferLoadSpec` sounds reasonable :)

fniephaus:
Mh, Gofer ships with Pharo and GemStone, but not with Squeak I think.
But all images come with Monticello, so I'd prefer a `SCIMonticelloLoadSpec` if that'd work...

fniephaus:
On second thought and since there already is a concept of platforms in smalltalkCI, we could add a `SCIGoferLoadSpec` for convenience.

PR #153, opened 2016-06-09
Add SCIMonticelloLoadSpec and SCIGoferLoadSpec

fniephaus:
This adds support for loading projects with Monticello and Gofer (#152).
Currently, it is only possible to load from a `MCHttpRepository`, but we should be able to support other MCRepositories (e.g. `MCFileTreeRepository`) in the future.

Please let me know what you think about this, @Uko.

Maybe @dalehenrich wants to have a look at this as well.


dalehenrich:
I'm curious how load order will be preserved if you mix #packages and #versions ... if gofer loads all of the packages at once I suppose it won't matter ... don't recall what the standard behavior of gofer is ... otherwise looks cool ...

fniephaus:
Currently, if you mix `#packages` and `#versions`, it would load all versions first and then the packages. I'm not sure if this is a problem, but I'm also not sure if mixing both is a good idea in the first place.
We either could merge the list into one (e.g. `#packagesAndVersions`) to preserve the order, or one could define more than one `SCIMonticelloLoadSpec` or `SCIGoferLoadSpec` (to avoid confusion, we could disable to mix `#packages` and `#versions`, but I'm again not sure if this would be better).

Uko:
I think that this makes sense. We have to see how people are going to use this to make improvements in the future

fniephaus:
Ok cool, just wanted to make sure that this is what you were looking for...

dalehenrich:
Well package load order is a thing, so mixing packages and versions would break the ability to control load order ... there are other ways to skin the load order cat, so this isn't a deal breaker.... and as @Uko mentions it's worth waiting to see how the usage evolves ...

PR #157, opened 2016-06-10
Add the ability to increase timeout for jobs with no output for 10mins

### LINKED ISSUE #154: Timeouts on long tests
Travis times out if there is no output for 10 minutes.

And naturally tests do not generate any output, so if they are running more than 10 min, the build is aborted. There is a documentation about this on https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

I think that it's a nice feature to abort builds, but maybe we can add some config that will allow to run the command prepended with `travis_wait` and a specified amount of time

fniephaus:
Closes #154.

/cc @Uko @dalehenrich

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ It is inspired by [builderCI][builderCI] and aims to provide a uniform and easy
 way to load and test Smalltalk projects.
 
 
-## Table of contents
+## Table Of Contents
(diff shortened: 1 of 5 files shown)


Uko:
Looks good

PR #162, opened 2016-06-13
Separate project loading and testing

fniephaus:
This is a preparation for deployments.

The build process now works as follows:
- Open image and load smalltalkCI
- Load project
- Close and save image
- Open image again
- Run project tests
- Save and quit if a test failed, otherwise just quit

This leaves us with a clean image when the tests pass, otherwise the image can be used for debugging purposes.

This PR also includes a few improvements (status code passing removed, GemStone package renamed)

Could you please have a quick look at this, @dalehenrich?


dalehenrich:
well I **am** concerned that you are (apparently) removing the old api: loadCI:, runCI:, and testCI:. I have code in my createClient and startClient scripts in GsDevKit_home that call the *old* methods and will break if you just up and remove them ... It's hard to tell from looking at the diffs in this pr, so if you aren't actually removing the methods then my concern is unfounded ...

fniephaus:
I have renamed the instance side methods to `load` and `test`. `GsDevKit_home` and your other scripts use the class side methods as entry points that I have adjusted accordingly. I only have cleaned up the ones that do not seem to be used by your projects anymore.

dalehenrich:
Well it's the "do not seem to be used" part that I am concerned about ... are you saying that the startClient failure was caused by a mistake here? If so I feel a little bit better ... and I do tend to agree that I pretty much stick to the class-side calls, so if those are preserved then I am less concerned :)

fniephaus:
Yes, I turned a method into an extension method for GemStone, but it was also used in Pharo clients for GemStone.

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/ensureAuthor.st
```diff
@@ -0,0 +1,4 @@
+initialize-release
+ensureAuthor
+	SmalltalkCI getAuthor ifEmpty: [
+		self class setAuthor: self class defaultAuthor ]
```
dalehenrich: It looks like you make the assumption that #getAuthor will return an empty string and that is not the case on GemStone ... again I can't tell, but the old way in prepareForCI looks like the author was preserved and right now I can't tell if the old way of preserving is still enforced

fniephaus: I have removed the preserving because it's not needed anymore. `SCIMetacelloLoadSpec` now only needs to make sure that the author is not empty, because that can cause loading or testing issues. Changing the build process as described in https://github.com/hpi-swa/smalltalkCI/pull/162#issue-159992343 makes all the `finishUpAfterCI` logic redundant.

fniephaus:
@dalehenrich If you don't have any objections, I'd like to merge this and I'll open pull requests to update `GsDevKit_home` and `GsDevKit_todeClient` accordingly. Am I missing any projects that directly depend on smalltalkCI?

dalehenrich:
Absolutely go ahead an merge ... let's see, I'm using .smalltalk.ston to create clients for the following projects:
- https://github.com/dalehenrich/Tugrik
- https://github.com/dalehenrich/voyage
- https://github.com/GsDevKit/GemStone-GCI

but they should all be using the "standard" createClient/startClient code in GsDevKit_home ... Currenlty they depend upon smalltalkCI directly to get the default stone name ... but that can change over time and needs ...

PR #175, opened 2016-07-06
Bugfixes for #172 and #174

### LINKED ISSUE #172: Use `startTopaz` instead of `devKitCommandLine` for loading and tests
[Metacello tests do a lot of logging and there are a lot of tests and the tests take much longer to complete than when run using builderCI ... theory is that the overhead for `Transcript` show, when it takes a round-trip to the client is way too high](https://github.com/dalehenrich/metacello-work/issues/390#issuecomment-219298570)..

### LINKED ISSUE #174: SCIMetacelloLoadSpec>>getClassesForMetacelloProjectRegistration:  can give wrong answers under certain circumstances
Metacello can have *conflicting* registrations for projects with the same basename. For example there can be registrations for **BaselineOfMetacello** and **ConfigurationOfMetacello** - typically this occurs in a [bootstrapped GemStone image](https://github.com/dalehenrich/metacello-work/issues/390#issuecomment-230639317), but can happen in Squeak and Pharo as well ...

The solution is to simply use the the full className (`BaselineOf` or `ConfigurationOf`). 
We can tell the difference between a basline and configuration in the call for [`SCIMetacelloLoadSpec>>getClassesForMetacelloProjectRegistration:`](https://github.com/hpi-swa/smalltalkCI/blob/master/repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/loadedClasses.st#L3-L6), but `SmalltalkCI class>>classesOfProjects:`  will take a little more research

dalehenrich:
Both of the following fixes have been tested with [recent Metacello builds](https://travis-ci.org/dalehenrich/metacello-work/builds/142914202) and are working as expected:
1. Fix #172: Use `startTopaz` instead of `devKitCommandLine` for loading and tests
2. Fix #174: SCIMetacelloLoadSpec>>getClassesForMetacelloProjectRegistration: can give wrong answers under certain circumstances

A bonus is that I found how to speed up the Metacello builds about 5x --- not related to SmalltalkCI (other than a bit for #172) --- the big was eliminating the unnecessary overhead for removing classes during Metacello tests ... being able to run SmalltalkCI in the tODE environment and using the profiling tools was the trick to speeding things up...


fniephaus:
Nice! I'm curious, how did you test the fixes? Locally or with [this](https://github.com/hpi-swa/smalltalkCI#using-a-different-smalltalkci-branch-or-fork)?

dalehenrich:
:) I pointed at the dkh-1 branch ... which is very convenient!

fniephaus:
So you used the `smalltalk_edge` facility, sweet! :)

PR #271, opened 2017-04-18
Fixed headings and anchors.

peteruhnak:
The anchor has to be inside the heading, otherwise it breaks the output


fniephaus:
Thanks, Peter! 😃

PR #292, opened 2017-07-05
Run Pharo images without default preferences

Uko:
Pharo users can have special preferences that run on image startup. This introduces unexpected issues when trying to reproduce Smalltalk-CI behavior locally.

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -233,7 +233,7 @@ pharo::run_script() {
     vm_flags="--no-quit"
   fi
 
-  travis_wait "${resolved_vm}" "${resolved_image}" eval ${vm_flags} "${script}"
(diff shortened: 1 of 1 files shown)


fniephaus:
I didn't know about `--no-default-preferences`, but seems to make sense. Is it supported by all Pharo VMs starting with the one for `Pharo-3.0`?

Uko:
Right, I didn't think back to 3.0 :). But the Travis build of SmalltalkCI runs on all the versions, no? So if it passes, the things didn't break. Otherwise I will check manually

fniephaus:
Yes, it seems to be working just fine. I was just wondering when this flag was introduced.

Uko:
I don't remember exactly, but I was using it for some years already

Issue #286, opened 2017-06-28
Support of 64bit Pharo image


Uko:
It would be nice to have a support for Pharo 64bit image & VM.

Image: http://files.pharo.org/get-files/60/pharo64.zip
VM:
 - http://files.pharo.org/get-files/60/pharo64-linux-threaded-stable.zip
 - http://files.pharo.org/get-files/60/pharo64-mac-stable.zip

The same applies to the 7.0 version

macta:
Yes - I've hit this one too

fniephaus:
What packages are required to run a Pharo 64bit vm on Precise/Trusty?

Uko:
Ah, right, there should be required libraries… I think that with 64 bits it has to be easier than 32. I will ask people who use Linux

fniephaus:
Looks like no additional packages need to be installed on Travis machines? :)

estebanlm:
yes, probably. 
maybe cairo but I think is already there.

Esteban

fniephaus:
Closed by 286e0f2e631418b5de7c20d161e1ee990be059b1.

Also, I can confirm that there's no need to install any additional dependencies on Travis to run an OpenSmalltalkVM/PharoVM :)

PR #295, opened 2017-07-07
Added 64 bit support

Uko:
I added a support for Pharo 64 bit images/vms as requested in the issue #286

Now I think that maybe it makes more sense to have one more matrix axis fir 32/64 architectures? Squeak seems to be experimenting with 64 bit so probably they will release a stable version in a near future too. And Pharo 64 bit scripts mostly mirror the 32 bit ones

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,11 @@ smalltalk:
   - Pharo-4.0
   - Pharo-3.0
 
+  - Pharo64-alpha
+  # - Pharo64-stable  # identical to Pharo64-6.0
+  - Pharo64-7.0
+  - Pharo64-6.0
+
   # - Moose-trunk   # identical to Moose-6.1
   - Moose-6.1
   - Moose-6.0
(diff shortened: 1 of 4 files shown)


fniephaus:
Thanks! This looks exciting :)
I would like to try to pollute the bash environments as little as possible with variables, so I'd prefer to keep the architecture selection in the image name. But it could go like this:
There could be `Pharo32-6.0` and `Pharo64-6.0`, while `Pharo-6.0` defaults to `Pharo32-6.0`, and in the future `Pharo-8.0` might default to `Pharo64-8.0`.

@dalehenrich @estebanlm: how do you like the `Pharo32-` and  `Pharo64-` prefixes? I'd like to make them consistent, so there would also be `Squeak64-`, `GemStone64-`, and so on.

Also, 32bit dependencies no longer need to be installed for 64bit vms, so I'll work on updating Travis once we decided on the naming convention. This should speed up 64bit builds a little bit...

Uko:
@fniephaus yes, I think you are right and the majority of developer will not really care about the 32 and 64 but rather use the recommended one. So I support the idea of having prefixes with architectures, and the ones without architecture that reference a recommended one.

PR #299, opened 2017-07-07
A better minimal .smalltalk.ston example

macta:
I think this is a more obvious minimal example - as not specifying any tests will automatically fail the build. The use of packages as a directory is also not the default that iceberg uses now - so ./ is more inline with what you get out of the box (you can leave it out, but this might be more obvious?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,10 +92,13 @@ SmalltalkCISpec {
   #loading : [
     SCIMetacelloLoadSpec {
       #baseline : 'MyProject',
-      #directory : 'packages',
+      #directory : './',
       #platforms : [ #squeak, #pharo, #gemstone ]
     }
   ]
+  #testing : {
+      #failOnZeroTests : false
+  }
 }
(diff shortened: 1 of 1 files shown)


Uko:
I like this suggestion. I don't know about default directory for code in other dialects, but Pharo tools definitely have `./` by default. Now regarding failing on no tests, I think that the minimal example should rely on default values. Do we really need to fail by default if there are no tests? How do the other Travis build systems work?

fniephaus:
Thanks for the suggestions. I noticed that Pharo tools use the root directory by default. However, I don't want to recommend `#directory : './'`, instead the key should be dropped. So let's further minify the minimal example ;)
With regard to `#failOnZeroTests`, this is an option for advanced users which I also wouldn't want to recommend. Some users didn't realize that their builds were always passing, because of an incorrect configuration. That's why `#failOnZeroTests` is enabled by default.

PR #300, opened 2017-07-10
Handle OCSemanticWarnings

fniephaus:
@dalehenrich can I have your blessing for [this](https://github.com/hpi-swa/smalltalkCI/compare/dev?expand=1#diff-df1d7d367d0662c6dd79e0177b241458) and [that](https://github.com/hpi-swa/smalltalkCI/compare/dev?expand=1#diff-e8f1bdaced291a98f28562de11333560) change?
Using a method on SmalltalkCI and solving this with inheritance seemed overkill here.
But it's more about handling `OCSemanticWarning`s in general. Is that ok?


fniephaus:
I'll fix the `Pharo64-alpha` build tonight...

Issue #293, opened 2017-07-05
Smalltalk CI not loading the current version present in gitfiletree repo


Uko:
The concrete case can be experienced when running SmalltalkCI on https://github.com/Uko/Renraku/tree/forced-rules

It supposedly loads the latest version of `Renraku-Test` because the logs have this line:
    
    Loaded -> Renraku-Test-cypress.1 --- filetree:///Users/uko/Downloads/Renraku-forced-rules [:] --- cache

but when I check the changes of Renraku-Test with the `filetree:///Users/uko/Downloads/Renraku-forced-rules` repo afterwards, it finds a lot of changes, and in fact, Renraku-Test is not updated to the latest version. I tried to debug this, but everything looks fine until I reach Metacello loading mechanism, and then it gets super complicated to debug.

Uko:
Ok, now I tried to manually load the changes from filesystem with
```
Metacello new
	repository: 'filetree://path/to/repo';
	baseline: 'Renraku';
	load
```
and I get a conflict while trying to merge. Can this be the issue? Can I force-load a package? I have no idea why there is a conflict, but maybe the metadataless format screws something up

fniephaus:
Have you tried with `#onWarningLog : true` or `#useLatestMetacello : true`?

See: https://github.com/hpi-swa/smalltalkCI#scimetacelloloadspec

Uko:
Yes, I've tried many options. Actually this one: https://travis-ci.org/Uko/Renraku/jobs/250653907 is with `#onWarningLog : true` and `#useLatestMetacello : true`. And I don't see any warnings in the log…

fniephaus:
Mh, interesting..will have a look soon...

Uko:
Looks like the packages are getting dirty during loading, and this causes problems. I'll try to change them to "unmodified" during a pro-commit hook and see if it works. But this looks like a strange MC behavior (or I used too much of a black magic somewhere).

theseion:
I've analyzed the problem. It's a combination of unfortunate logic in Monticello and exception handling in SmalltalkCI. @fniephaus, you can mark this issue as bug and I'll provide a PR with a fix and complete description of the problem.

theseion:
See https://github.com/pharo-project/pharo/pull/163 for the fix to Monticello.

(@fniephaus I don't have permission to alter an issue, you'll have to change the tag from "question" to "bug" yourself. Apparently, only maintainers are aloud to do that.)

[...]

PR #302, opened 2017-07-10
Correctly resume MCMergeOrLoadWarning in Pharo

theseion:
* refactored load warning exception handlers
* added handler for MCMergeOrLoadWarning

fixes issue #293

[ci skip]


fniephaus:
Mh, this broke the GemStone builds ([see here for example](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/252285938)).

theseion:
I used `#at:ifPresent:ifAbsent:` to look up `MCMergeOrLoadWarning`. I'll have to use a more conservative method then. I'll fix it tonight.

fniephaus:
Maybe solve this via inheritance? If this is only needed for Pharo, add a class method to `SmalltalkCI` and override it in `SmalltalkCIPharo`.

theseion:
Sounds sensible.

theseion:
I've pushed the fix for this to dev directly. BTW, I opted to create subclasses of `SCIMetacelloLoadSpec` (as already exist for `SCIMonticelloLoadSpec`) instead of using a class side method.

PR #311, opened 2017-07-25
add Pharo 6.1 and a couple of env vars we may use

estebanlm:


estebanlm:
do you know why this is failing @fniephaus ?

pharo/run.sh
```diff
@@ -87,9 +93,45 @@ lastSuccessfulBuild/artifact/${moose_name}.zip"
...
+# variables: 
+#  PHARO_VM=stable*|latest
+#  LINUX_HEARTBEAT=threaded*|itimer
 pharo::get_vm_url() {
   local smalltalk_name=$1
+  local os="$(pharo::get_os)"
+  local heartbeat=""
+  local latest=""
 
+  if [ "${PHARO_VM}" == "latest" ]; then
```
fniephaus: Needs to be `${PHARO_VM:-}`, otherwise you get the `PHARO_VM: unbound variable` error.

README.md
```diff
@@ -506,6 +514,17 @@ This allows to identify build errors caused by smalltalkCI updates and
...
 `--no-tracking` option.
 
+### Pharo-specific Options
+
+Pharo added some specific options to customise its builds: 
+
+```yml
+env:
+  global:
+  - PHARO_VM=stable|latest              # choose stable/latest VM (default: stable)
```
fniephaus: We will soon have `smalltalk_vm` support in Travis.
Instead of adding Pharo-specific env vars, can we do something like the follow?
```yaml
smalltalk_vm:
  - Pharo-7.0-stable-threaded
  - Pharo64-6.0-latest-itimer
```

pharo/run.sh
estebanlm: mmm, ok :)

README.md
estebanlm: that will create a lot of possibilities for each version of Pharo (not to say that the distinction between itimer and threaded are just available to linux).
honestly, for the moment I would prefer to keep it this way... like "extras" you can add...

fniephaus: Hi @estebanlm,
We've tried hard so far to stay away from adding dialect-specific options via `env`. The main reason is that we'd need to support them forever as soon as users start using them. That's why I am hesitating here.

Also, I don't understand how your proposed approach solves the problem with regard to itimer/threaded only being available on Linux. Setting `env: global: LINUX_HEARTBEAT=threaded` would also affect Mac builds, but it would be ignored.

I'm really just looking for the best (and generic) way to add support for manual vm selection and I still imagine something like this at the moment:
``` yml
jobs:
  include:
    - smalltalk: Pharo-7.0
      smalltalk_vm: Pharo-7.0+latest
    - smalltalk: Pharo64-7.0
      smalltalk_vm: Pharo64-7.0+stable+itimer
    - smalltalk: Squeak-5.1
    - smalltalk: Squeak-5.1
      smalltalk_vm: OpenSmalltalkVM+latest
```
Everything with `+` are vm options, so `Pharo-7.0+stable+itimer` would be the same as `Pharo-7.0+itimer+stable`.

fniephaus: One more example to demonstrate why I'm trying to avoid dialect-specific env vars.
At some point, we want to be able to select latest VMs in Squeak and Moose as well, and then we end up with something like this:
```yml
env:
  global:
  - PHARO_VM=latest
  - SQUEAK_VM=latest
  - MOOSE_VM=latest
```

Instead, we could even support this:
```yml
smalltalk_vm: +latest
```
This would automatically select the right VM for the right image (Pharo, Squeak, Moose) with the `latest` option enabled, so basically the same as writing `Pharo-X+latest`,  `Squeak-X+latest`, and  `Moose-X+latest`.

Issue #306, opened 2017-07-14
Projects in GitHub won't load in Pharo 7 on Travis


gcotelli:
Hi,

I'm having this error:

```
Loading baseline of BaselineOfWillow...Creating iceberg-metacello adapter for: github://ba-st/RenoirSt:master/source

LGit_GIT_ERROR: error authenticating: no auth sock variable
```
Check the full log output in:
https://travis-ci.org/ba-st/Willow/jobs/253572685

The build works Ok por Pharo 5/6 , but fails on Pharo 7 (running in Linux)

theseion:
This is probably an Iceberg / LibGit2 problem, not specific to smalltalkCI. Can you reproduce the problem when you evaluate the load instructions manually?

gcotelli:
Works ok in my local machine with a recently downloaded image and VM:

Image
-----
/home/gcotelli/Pharo/Pharo7-32/Pharo.image
Pharo7.0SNAPSHOT
Latest update: #0
Unnamed

Virtual Machine
---------------
/home/gcotelli/Pharo/Pharo7-32/pharo-vm/lib/pharo/5.0-201707132359/pharo
CoInterpreter VMMaker.oscog-eem.2253 uuid: 45202565-6d57-4e4b-812f-a8eea4685e21 Jul 14 2017
StackToRegisterMappingCogit VMMaker.oscog-eem.2252 uuid: 2f3e9b0e-ecd3-4adf-b092-cce2e2587a5c Jul 14 2017
VM: 201707132359 https://github.com/OpenSmalltalk/opensmalltalk-vm.git $ Date: Thu Jul 13 16:59:42 2017 -0700 $ Plugins: 201707132359 https://github.com/OpenSmalltalk/opensmalltalk-vm.git $

Unix built on Jul 14 2017 00:14:07 Compiler: 4.6.3
VMMaker versionString VM: 201707132359 https://github.com/OpenSmalltalk/opensmalltalk-vm.git $ Date: Thu Jul 13 16:59:42 2017 -0700 $ Plugins: 201707132359 https://github.com/OpenSmalltalk/opensmalltalk-vm.git $
CoInterpreter VMMaker.oscog-eem.2253 uuid: 45202565-6d57-4e4b-812f-a8eea4685e21 Jul 14 2017
StackToRegisterMappingCogit VMMaker.oscog-eem.2252 uuid: 2f3e9b0e-ecd3-4adf-b092-cce2e2587a5c Jul 14 2017

[...]

PR #312, opened 2017-08-01
Fix for #306

fniephaus:


fniephaus:
@theseion why do you always use `[ci skip]` in your commit messages?

gcotelli:
The Moose builds are failing due to INRIA infrastructure. There was a problem with the Jenkins setup.

theseion:
Because you say so in the readme. But maybe you were only referring to PR's?

fniephaus:
Oh, that comment is only relevant when adding a project to the `README.md`. No need to spin up all builds just because the `README.md` is changed.
`[ci skip]` is good if you can guarantee that the commit does not change smalltalkCI's logic/behavior.

theseion:
Makes sense :)

PR #322, opened 2017-10-18
Gitlab coverage

peteruhnak:
Coveralls doesn't support GitLab, however I would like to use the existing infrastructure/export for coveralls.

`aws s3 cp "${SMALLTALK_CI_BUILD}/coveralls_results.json" s3://$S3_BUCKET/$CI_PROJECT_PATH/$CI_JOB_ID/coveralls_results.json`)


peteruhnak:
And considering a lot of metadata is being removed... I guess I should recommit it using metadata-full FileTree (which I'll do as soon as I figure out how)

PR #323, opened 2017-10-19
Gitlab coverage

peteruhnak:
Coveralls doesn't support GitLab CI, however I would like to use the existing infrastructure and the coveralls exports.

aws s3 cp "${SMALLTALK_CI_BUILD}/coveralls_results.json" s3://$S3_BUCKET/$CI_PROJECT_PATH/$CI_JOB_ID/coveralls_results.json)

apparently a lot of metadata were updated to reflect the last author...


fniephaus:
Hi @peteruhnak,
Could you please make sure to only touch the metadata related to your change? This would make it easier to review your change...

peteruhnak:
Yeah, I am trying to figure out what I should use so other metadata is not touched... Iceberg doesn't include them, GitFileTree stores them incorrectly, and FileTree loads them incorrectly.. so I am not sure. 🤔

fniephaus:
Maybe use an older version of Pharo with metadata enabled? smalltalkCI still needs metadata at this point to provide compatibility for older images that do not support metadata-less FileTree yet..

estebanlm:
Pharo still support metadata files. But you need to add them as simple "filetree" repositories :)

PR #324, opened 2017-10-19
Enabled coverage testing (coveralls export) for GitLab CI

peteruhnak:
Coveralls doesn't support GitLab CI, however I would like to use the existing infrastructure and the coveralls exports.

aws s3 cp "${SMALLTALK_CI_BUILD}/coveralls_results.json" s3://$S3_BUCKET/$CI_PROJECT_PATH/$CI_JOB_ID/coveralls_results.json)

So the way I had to commit it was...
1) for SmalltalkCI-Core package load it with Iceberg and commit with FileTree (in P6)
2) for SmalltalkCI-Pharo-Tests package load it with FileTree and commit with FileTree (in P5)

Which makes me think that different people in the past used different methods to commit the code, so using the same method across packages will mangle the metadata.


helpers.sh
```diff
@@ -322,6 +341,10 @@ upload_coverage_results() {
   local curl_status=0
   local coverage_results="${SMALLTALK_CI_BUILD}/coveralls_results.json"
 
+  if is_gitlabci_build; then
+    return 0 # Uploading results is not supported for GitLab CI
```
fniephaus: So you don't actually upload the file here, but you do that separately at a later point in time? Basically, you just want the coverage result file? If this is the case, it'd be confusing to claim to support coverage for GitLab. Maybe we rather need an option to export coverage, even when not running on Travis/AppVeyor.

peteruhnak: Yes, as I've mentioned in the Pull Request -- Coveralls doesn't support GitLab, but I want to use the export file.

fniephaus: Mh, in that case, I'd prefer to have a boolean option in [`#coverage`](https://github.com/hpi-swa/smalltalkCI#coverage-testing) like `#forceExport: true`. This would be independent from GitLab and would also avoid to export the results for GitLab projects that don't actually make use of it. Makes sense?

peteruhnak: Makes sense to me, I'll update it.

peteruhnak: One more thing. Testing whether coverage should be ran is tested three times:
1) first here in `export_coveralls_data`
2) second in `SmalltalkCI>>isCoverageTestingEnabled` (called from `SmalltalkCI>>test`)
3) third in `SCICodeCoverage>>run:spec:in:` where it checks whether the coveralls data file (from step 1)) was created

I suppose I need to check the `#forceExport:` in `isCoverageTestingEnabled` too, or we can base it only on the existence of the file.

fniephaus: How about this:
- rename `export_coveralls_data` to `export_coverage_data` (and the actual file) and move [this check](https://github.com/hpi-swa/smalltalkCI/blob/b35f3ec6f4baa41c34e17e5396f098dd6eb112f2/helpers.sh#L288-L294) to the [`upload_coverage_results `](https://github.com/hpi-swa/smalltalkCI/blob/b35f3ec6f4baa41c34e17e5396f098dd6eb112f2/helpers.sh#L321)
- [`SCICodeCoverage>>run:spec:in:`](https://github.com/hpi-swa/smalltalkCI/blob/354bf5eaa6c1d7ea81af69bfd1ec6066bdd11c17/repository/SmalltalkCI-Core.package/SCICodeCoverage.class/instance/run.spec.in..st) should work as is, maybe rename coveralls to coverage as well if you have time
- [`SmalltalkCI>>isCoverageTestingEnabled`](https://github.com/hpi-swa/smalltalkCI/blob/946ca1c0c2d808cb093a15392a38959a91011c03/repository/SmalltalkCI-Core.package/SmalltalkCI.class/instance/isCoverageTestingEnabled.st) is the method that needs to check for the new option (because you can't check that flag in Bash, e.g. in `export_coverage_data`, because there's no STON parser for Bash).

Hope this still makes sense :)

peteruhnak: Regarding renaming coveralls -> coverage. Technically we are still exporting coveralls data, so I would vote for keeping the name.

peteruhnak: In fact `upload_coverage_results` should be renamed to `upload_coveralls_results`.
Regarding the check there, it is already checked in `run.sh` ( https://github.com/hpi-swa/smalltalkCI/blob/eca514d90f190b4692fb0ff311c48e00e97e1976/run.sh#L621 ). But we can duplicate it to make sure? (Or remove it from run.sh)

fniephaus: Keep the check in run.sh and remove it in the other file. upload_coveralls_results sounds good, consistency is more important :)

peteruhnak: Do we really need option `forceExport:`? Doesn't the fact that I provide the coverage info in the spec file imply that I want to run the coverage?
https://github.com/hpi-swa/smalltalkCI/blob/eca514d90f190b4692fb0ff311c48e00e97e1976/repository/SmalltalkCI-Core.package/SmalltalkCISpec.class/instance/coverageEnabled.st#L1-L3

fniephaus: Agreed, no need to add `forceExport`, always export when the config contains `#coverage` :)

repository/SmalltalkCI-Core.package/SmalltalkCI.class/instance/isCoverageTestingEnabled.st
```diff
@@ -1,5 +1,3 @@
 accessing
 isCoverageTestingEnabled
-	^ SmalltalkCI platformSymbol ~= #gemstone and:
-		[ (SmalltalkCI isTravisBuild or: [ SmalltalkCI isAppVeyorBuild ]) and:
-			[ self spec coverageEnabled ] ]
\ No newline at end of file
+	^ SmalltalkCI platformSymbol ~= #gemstone and: [ self spec coverageEnabled ]
```
peteruhnak: I've removed the extra travis/appveyor checks too, as they are no longer needed (if we consider has coverage -> we want export)

helpers.sh
```diff
@@ -282,21 +286,34 @@ git_log() {
   echo "${output/\"/\\\"}" # Escape double quotes
 }
 
+
 export_coveralls_data() {
-  local service_name
+  local service_name=""
```
peteruhnak: If it is an unknown environment (e.g. local) then we just leave the metadata empty.

fniephaus: Maybe better to actually use `local service_name="unknown"`.

fniephaus: Same with the below (`branch_name`, etc)

peteruhnak: changed

Issue #333, opened 2017-11-06
Handling load conflicts in a SCIMetacelloLoadSpec (and other Metacello Scripting API functions)


jbrichau:
I'm moving a project to Travis-CI where the Metacello Scripting API is required to handle load conflicts in an `onConflict:` handler. The project's baseline references many other project configurations and we need to handle several load conflicts (e.g. different projects reference different versions of Seaside) in order to load the code into a Pharo image.

We currently do that using a custom Metacello load script implemented on the class side of our BaselineOf class. This custom script also sets a `repositoryOverrides:` because we replicate all dependent packages to a folder in the same git repository. This makes our load independent from availability or changes in outside repositories.

Is there a way to do this in Smalltalk-CI ? I have been experimenting with the `#preLoading` script, which allows me to load the project but it seems I need to add a `#loading` declaration as well or it will not work. I think I can make a stub load group to make this work but it feels like a workaround, while I would actually want to use additional functions of the Metacello API.

If this is not available, I'm happy to take pointers to how this would be possible to implement and give a PR a go. We need to move this project to Travis-CI, so definitely interested to extend Smalltalk-CI if possible.

fniephaus:
Hi @jbrichau,
Could you please provide such a Metacello load script for you project?
The `#preLoading` option is available since last week, so maybe it's a bug that `#loading` is required. I believe it shouldn't be required.
How do your users install your project? Do you make them handle the conflicts themselves? Or could you avoid any conflicts in a `preLoad` script in your baseline?

jbrichau:
Hey @fniephaus,

The project is a private project (Yesplan), so I cannot share the entire config out-of-the-box. 
Here is the load script (which is a class method of our BaselineOf). Do not mind the exact filepath in there, I am experimenting to make this work :):

```Smalltalk
loadDevelopment
	"self loadDevelopment"
	| repository |
	repository := MCFileTreeRepository new directory: (FileSystem workingDirectory / 'git_cache' / 'repository').
	Metacello new
   		baseline: 'NextPlan';
   		repository: repository description;
		repositoryOverrides: (self otherRepositories add: repository description; yourself);
		onConflict: [:ex | 
    				(ex newProjectRegistration projectName = ex existingProjectRegistration projectName and:[ex newProjectRegistration version = ex existingProjectRegistration version]) 
					ifTrue: [ ex allow]
    				ifFalse:[ 
[...]

PR #335, opened 2017-11-11
Support custom scripts in the loading phase 

jbrichau:
Changes to support a custom script in the loading phase. See issue #333 for discussion.

This PR is not yet finished I think (i.e. it was only tested for our own builds). 
I took the option to extend the `SCICustomScript` with the interface required by a loading spec because it's convenient to use the implementation of scripts in that loading phase. Would a different option suite the architecture better?


fniephaus:
Thanks, I'll get that merged into `dev` first. There's a couple of things I have to take care of first...

fniephaus:
Squashed and merged into `dev` via https://github.com/hpi-swa/smalltalkCI/commit/fceeac872e4a711a9feeec9eef4fc120ece69eea.

PR #345, opened 2017-11-22
abstract methods might not revelant lines of code

mattonem:
Abstract methods might not be revelant to be tested.
At least I never test abstract methods.

### DIFF
--- a/repository/SmalltalkCI-Core.package/SCICodeCoverage.class/instance/coverageDictFor.in..st
+++ b/repository/SmalltalkCI-Core.package/SCICodeCoverage.class/instance/coverageDictFor.in..st
@@ -2,7 +2,11 @@ helpers
 coverageDictFor: aCompiledMethod in: aDirectoryName
 	| filePath coverageSymbol |
 	filePath := (self filePathFor: aCompiledMethod in: aDirectoryName).
-	coverageSymbol := aCompiledMethod hasRun ifTrue: [ 1 ] ifFalse: [ 0 ].
+	coverageSymbol := aCompiledMethod hasRun
+		ifTrue: [ 1 ]
+		ifFalse: [ aCompiledMethod isAbstract
+				ifTrue: [ nil ]
+				ifFalse: [ 0 ] ].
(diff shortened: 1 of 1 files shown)


fniephaus:
Wouldn't it be better to ignore abstract methods before the code coverage is tested, e.g. in [here](https://github.com/hpi-swa/smalltalkCI/blob/289c818c4797b63cd7c13846b0dde8aced6dfebe/repository/SmalltalkCI-Core.package/SCICodeCoverage.class/class/methodReferencesFor..st).

mattonem:
If you ignore the method before testing, it wouldn't be reported in coveralls as "not revelant" code line, i guess.
The null value is acceptable value. And it would be good to use it.

from http://docs.coveralls.io/api-reference: 
coverage | Array REQUIRED The coverage data for this file for the file’s job.The item at index 0 represents the coverage for line 1 of the source code.Acceptable values in the array:
- A positive integer if the line is covered, representing the number of times the line is hit during the test suite.
- 0 if the line is not covered by the test suite.
- null to indicate the line is not relevant to code coverage (it may be whitespace or a comment).

fniephaus:
So, ultimately this would lead to the same result on Coveralls? I mean not including a method file or including it with `null`?
The Smalltalk test runner rejects abstract methods before the coverage test is performed and I'd like to replicate that behavior as much as we can...

mattonem:
Except, if you put it with null you can still see it in coveralls, with revelant line value to 0. 
The abstract method is part of the software, I dont think it should just vanished from reports, but appear as unrevelant. 
like this:
https://coveralls.io/builds/14320787/source?filename=repository%2FSmalltalkCI-Core.package%2FSCICodeCoverage.class%2Finstance%2FreadCoverallsData.st

fniephaus:
Had to refactor some additional code, but basically it should now do the same thing.
Fixed with https://github.com/hpi-swa/smalltalkCI/commit/9d4e4921b8cf5faa887b119426ceb3c2079a85ad.

PR #353, opened 2018-01-25
pharo: detecting repository format (tonel/filetree)

estebanlm:
it will detect if we are using tonel or filetree. 
this will load tonel baselines, but if those baselines depends on other baselines on tonel, it will still fail (that requires a change in metacello I plan to do tomorrow)


fniephaus:
I know it's because of Iceberg, but it's unfortunate that the PR touches so many metadata files :(
I wish this was easy to fix....

dalehenrich:
IOt looks a bunch of the metadata file changes are due to the fact that the values in the ston files are not being sorted by pharo --- which is yet another bug ... I thought that bug would have been fixed by now ...

dalehenrich:
@estebanlm I'm curious about why Metacello would need to be changed to "detect filetree/tonel" repositories ...  IIRC there is an unresolved discussion on the Metacello group about what should be done, so I hope you don't make any changes to Metacello without discussing your plans with me...

estebanlm:
@fniephaus is not because of iceberg (I do not use it when working on this "metadata-full" formats), is because pharo use the internal STON package to generate the metadata files and that is not deterministic. 
As @dalehenrich says, that's "another bug" :P

estebanlm:
@dalehenrich I'm not making any change without discussing. What I did with this was simply adapting what was before to understand also tonel (before we had filetree:// and I make it understand tonel://). 
Since both are independent packaging formats, the different indication seemed reasonable. 
Now, in metacello I was not planning anything different fro what I understood of that thread you mentioned... but I will not touch anything without your agreement, as I always do.

PR #358, opened 2018-03-04
[Feature]: Docker

seandenigris:
Here is a crack at Docker support. There are basic implementations of the following:
- Programmatically Create Dockerfiles
  - Some of the API has been wrapped (e.g. RUN, FROM, and comments)
  - [An example Ubuntu prepped for Pharo 32-bit VM](https://github.com/seandenigris/smalltalkCI/blob/docker/docker/Dockerfile) is included in the commit under `docker`, based off of [Herby's](https://lolg.it/herby/pharo-docker/src/master/base_32/Dockerfile).  There is [a Docker Hub auto-build](https://hub.docker.com/r/seandenigris/pharo/) for it which can be used in CI jobs.
  - The barebones Pharo process is reified via SCIDockerPharo
- There are some tests

NB: the `-Docker` package has Magritte and GT extensions, which could be made optional as they were (so far) only used for playing during development, but I wasn't sure the best/easiest way to package them.


fniephaus:
Hi @seandenigris,
Thanks for the PR. I haven't worked much with Docker yet, so sorry for the dump question: how is this adding Docker support? The PR allows Dockerfiles to be generated, but who triggers this generator (dev or CI?) and how are the Dockerfiles being executed? Don't get me wrong, it'd be great if smalltalkCI can support Docker, I just would like to understand *how*. :)

seandenigris:
@fniephaus Great questions. I didn't know anything about Docker before this week either, ha ha. 
> The PR allows Dockerfiles to be generated, but who triggers this generator (dev or CI?)
Before committing ST code which adds/enhances Dockerfiles, the dev should generate any needed updates
> how are the Dockerfiles being executed?
Docker Hub will automatically pick up the new Dockerfiles when they are committed to GH and generate images usable in e.g. Travis and GL Pipelines
> how is this adding Docker support? 
Given the above, all there is left to do is define Dockerfile(s) for all the platforms we support. I included an example that sets up Ubuntu LTS for Pharo 6.1 with a 32-bit VM.

fniephaus:
Hi @seandenigris, sorry for not responding to this PR for so long.

While the idea of generating `Dockerfile` for smalltalkCI is interesting, I don't think the code for doing this should be part of the `master` branch. After all, all code on `master` will be pulled by *every* CI build and this code will never be needed directly for a build. Fortunately, the code is already persisted on GitHub. If you want, we could move it to a dedicated branch of this repo or feel free to maintain it in a separate repository. I hope you don't mind that I'm closing this PR for now.

BTW: a while ago I worked on a [Docker image for smalltalkCI](https://github.com/hpi-swa/smalltalkCI/tree/docker), which I believe should still work.

Happy to discuss if you have any questions or ideas!

seandenigris:
Thanks for taking a look, @fniephaus. I see what you mean. Let me think about a different approach

PR #361, opened 2018-03-21
Fix #360

### LINKED ISSUE #360: Should update Pharo Stdout access
There was some renaming in the classes providing access to stdio in latest dev Pharo 7. The  platform code should be fixed accordingly.

guillep:
Rename ZnCrPortableWriteStream into ZnNewLineWriterStream


fniephaus:
Does this need to be backwards compatible?

guillep:
Nope, this is for Pharo-alpha. The class that was renamed does not exist in Pharo<7. It was introduced in Pharo7 and now renamed.

fniephaus:
Ok, thanks!

fniephaus:
The change should now be live. Thanks again for the fix!

guillep:
Thank you!

PR #369, opened 2018-04-18
Fixed coverage reporting when a class is included but has 0% of coverage

### LINKED ISSUE #370: Coverage reporting failed when a class is included but has 0% of coverage

gcotelli:
Fixes #370


fniephaus:
Thanks!

PR #374, opened 2018-06-19
Supporting parametrized tests that produce a custom suite.

tesonep:
Pharo allows the subclasses of TestCase to generate a TestSuite with the tests in the TestCase. 
This allows us to create a parametrized test, one test configured with many options spawning different tests. 

Changing the building of the test suite to use #suite and then flattening this suites into the one used by SmalltalkCI.


fniephaus:
@estebanlm could you have a quick look at this?

estebanlm:
yes, seems ok

fniephaus:
Thanks to both of you, I should have time to merge this tomorrow!

fniephaus:
Thanks for the contribution, @tesonep!

PR #385, opened 2018-07-27
Allow symlinking the main script

cdlm:
Since bin/smalltalkci tries to auto-detect the installation prefix from its own path, it
did not work through a symbolic link. Resolving the symlink as a first step fixes that.


fniephaus:
Thanks, will have a look at this soon!

fniephaus:
Manually merged into `master` via https://github.com/hpi-swa/smalltalkCI/commit/bbf88508eea315daf949e83207d11e49d8fe42e2. Thanks for the contribution :)

PR #389, opened 2018-08-20
failOnZeroTests mixup

codeZeilen:
Besides the issue fixed in the commit there is another, maybe larger, one. When I ran a job with zero tests it initially failed which confused me as I always read the example configuration in the Readme.md as the standard configuration and so I was expecting the flag to be set to false and the build to not fail. Maybe we should add a sentence before the example configuration illustrating that these are merely some example values.

### DIFF
--- a/repository/SmalltalkCI-Core.package/SCITestRunner.class/instance/tearDown.st
+++ b/repository/SmalltalkCI-Core.package/SCITestRunner.class/instance/tearDown.st
@@ -8,4 +8,4 @@ tearDown
 	(self spec failOnZeroTests and: [ self totalTests == 0 ])
 		ifTrue: [ SCIError signal: 'No tests were executed.
 
-If this is intended, use `#failOnZeroTests : true` in your SmalltalkCISpec.' ].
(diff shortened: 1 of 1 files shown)


fniephaus:
IIRC, the values in the README.md are default values actually. Anyway, there should be some short explaination about those values.

codeZeilen:
Then we also have to change this method as it currently returns true by default:

https://github.com/hpi-swa/smalltalkCI/blob/master/repository/SmalltalkCI-Core.package/SmalltalkCISpec.class/instance/failOnZeroTests.st

Would that be the right way forward?

fniephaus:
The default behavior should be *to fail* on zero tests, otherwise users might think everything is ok when using a bogus config that has an empty set of classes for testing. This PR corrects the additional hint you would see in these scenarios, so I think this is good as it.

I'm afraid your confusion has confused me now a little bit too :D Is there still something unclear? The `README.md` does not contain defaults apparently. But I think `failOnZeroTests` is still descriptive enough for this option, right?

PR #393, opened 2018-09-11
Option to disable serialization of failing/erroring tests (e.g. with Fuel)

jbrichau:
In some of our builds, we hit errors in Fuel that made the entire build crash. 
In such cases it's useful to be able to disable the serialization of the error (using Fuel in Pharo) to get the build to work.

I added the option to pass this via the configuration, with the default being false (i.e. keep serialization on when the option is not provided).


jbrichau:
Ouch, I accidentally added commits to the branch that were not supposed to go into the PR. I realized to late and will fix.

jbrichau:
Ouch, I accidentally added commits to the branch that were not supposed to go into the PR. I realized too late and will fix.

jbrichau:
I reverted the commit.
This PR should be complete now :)

fniephaus:
Hi @jbrichau,
Thanks for the PR.
Two remarks:
- for consistency reasons, can we call the option "#serializeError"?
- maybe it makes sense to disable serialization by default? What does @theseion think?

jbrichau:
Hey @fniephaus,

I was hesitating myself between "serializeError" and "disableSerializeError", so no problem renaming it. Done!
I would say we need to keep the current behavior the default (i.e. keep the serialization) but having the option to disable it is perfectly fine for those who need/prefer it.

cheers

Issue #396, opened 2018-09-20
Need better way to set `GEM_TEMPOBJ_CACHE_SIZE` for GemStone/S


dalehenrich:
Right now the run.sh for GemStone uses the `-T 100000` option when starting topaz but occasionally [a user needs a larger TOC](http://forum.world.st/TravisCI-build-run-out-of-memory-td5084889.html) .... 

A couple of options here:
1. topaz is started twice during the run. The first time the SmalltalkCI and the project are loaded ... the second time, the tests are run ... if we arrange to recognize that a [gem.conf file is specified](https://github.com/hpi-swa/smalltalkCI/tree/master/gemstone#smalltalkci-scigemstoneserverconfigspec) then we could skip using the `-T 100000` option when we run the tests ... doesn't help if we need to bump up the TOC to load the projects
2. add a new env var/command line option for GemStone (GSCI_TOC and --gs-TOC) where the TOC size can be directly specified ... the TOC would be used for both topaz invocations ... solving the problem from option 1.

I think the second option is probably the best solution ...

fniephaus:
+1 for option 2 :)

fniephaus:
Reopening this (see #397).

PR #397, opened 2018-09-21
Update run.sh for GemStone

brunobuzzi:
Partially fix:
https://github.com/hpi-swa/smalltalkCI/issues/396
In startTopaz command after -T option ${GSCI_TOC:-100000}
It will force TOC size to get value from env variable GSCI_TOC if not present then the default value is used (100000)

This commit does not fix --gs-TOC option of #396 (i think is a tODE option).

### DIFF
--- a/gemstone/run.sh
+++ b/gemstone/run.sh
@@ -204,7 +204,7 @@ gemstone::load_project() {
   local status=0
 
   fold_start load_server_project "Loading server project..."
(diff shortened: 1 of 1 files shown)


dalehenrich:
The `--gs-TOC` option of #396 is not for tODE, but for running smalltalkCI from the command line, so that you are not forced to define an environment variable

PR #398, opened 2018-09-28
add #onConflict:/#onUpgrade: metacello logic

### LINKED ISSUE #333: Handling load conflicts in a SCIMetacelloLoadSpec (and other Metacello Scripting API functions)

estebanlm:
fixes #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,6 +383,8 @@ SCIMetacelloLoadSpec {
   #directory : 'packages',                            // Path to packages if FileTree is used
   #repository : 'http://smalltalkhub.com/mc/...',     // Alternatively, define MC repository
   #onWarningLog : true,                               // Log Warning messages to Transcript
+  #onConflict : #useIncoming,                         // When there is a conflict between loaded sources and incoming sources (can be #useIncoming|#useLoaded)
+  #onUpgrade : #useIncoming,                          // When loaded sources are an older version than incoming sources (can be #useIncoming|#useLoaded)
   #failOn : [ #OCUndeclaredVariableWarning ],         // Fail build on provided list of Warnings
   #useLatestMetacello : true,                         // Upgrade Metacello before loading
   #load : [ 'default' ],                              // Define MC load attributes


estebanlm:
the error on GemStone seems unrelated.

fniephaus:
Thanks, @estebanlm!
@dalehenrich could you please have a look at this PR, should be relatively quick...I'm not too familiar with this part of the Metacello API, but you should since you designed it 😅. Thanks!

estebanlm:
It just exposes the api already existing so I don’t think there is a problem :)

fniephaus:
Ok, that makes sense then. I don't want to block you, but I'm currently traveling without a laptop. Will have a look and merge it next week if that's ok.

estebanlm:
Yes. I think for complex cases is better to use a preLoad script, since is hard to model that into a ston file.

fniephaus:
@estebanlm I have merged (and cleaned up this PR) in 3dc699f560c8d4b512b40a7bae7d4e8631b070c7. Could you please have a look at https://github.com/hpi-swa/smalltalkCI/commit/3dc699f560c8d4b512b40a7bae7d4e8631b070c7#diff-df1d7d367d0662c6dd79e0177b241458R16? I think it's better to check that `#onConflict` and `#onUpgrade` are not `nil` (which they are when not set in a `smalltalk.ston`)...

estebanlm:
well, is about philosophies. I preferred the other way (explicit default instead inferred with nil) :)
but is ok, as long as it is integrated. 

btw... what's the pattern to move dev to master?

fniephaus:
> well, is about philosophies. I preferred the other way (explicit default instead inferred with nil) :)
but is ok, as long as it is integrated.

Well, if any of these two values is `nil`, wouldn't `onConflict: [ :ex :loaded :incoming | ex perform: nil ]]` throw an exception? The other option is to lazy-initialize the values or provide default return values in the getters...

> btw... what's the pattern to move dev to master?

I merged commits onto `dev` first, everything that is merged is supposed to reach `master` (so nothing experimental). Only if `dev` passes, it's merged into `master` to avoid breaking any user-facing builds. Hope that makes sense :)

fniephaus:
https://github.com/hpi-swa/smalltalkCI/commit/3dc699f560c8d4b512b40a7bae7d4e8631b070c7 is now merged into `master`.

estebanlm:
no, because I was doing 

```Smalltalk
metacello onConflict: [ :ex :loaded :incoming | ex perform: self onConflict ]]
```
and onConflict was defined like this: 

```Smalltalk
onConflict
    ^ onConflict ifNil: [ #useLoaded ].
```

but is ok ;)

fniephaus:
No, I'm afraid both [`onConflict.st`](https://github.com/estebanlm/smalltalkCI/blob/bc2d3aee0cce552854a934dc8ff83725d3f3b8c1/repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/onConflict.st) as well as [`onUpgrade.st`](https://github.com/estebanlm/smalltalkCI/blob/bc2d3aee0cce552854a934dc8ff83725d3f3b8c1/repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/onUpgrade.st) were just standard getters which is what concerned me in the first place. Anyway, it should work now :)

estebanlm:
then there was a bug while saving the package, and you are right it needs to be safe guarded :P

dalehenrich:
There was quite a bit of noise (changed property order) in the diffs, so it was not easy to see what was actually done ...  though that the order issue had been fixed:)

estebanlm:
me too. But seems it is not :(

dalehenrich:
haha ... I hate those bugs that don't stop bugging!

fniephaus:
> There was quite a bit of noise (changed property order) in the diffs, so it was not easy to see what was actually done ... though that the order issue had been fixed:)

I usually revert all property files that have been reordered during the merge only so that I can see what has actually changed. Maybe we should backport metadataless support for all images supported by smalltalkCI, so that smalltalkCI can finally be metadataless too and we can edit things directly here in GitHub.

dalehenrich:
@fniephaus, the property file noise is not a side-effect of metadataless mode ... it is because STON does not order the keys in dictionaries by default and for tonel and filetree, the writers need to sort the keys alphabetically ... at some point FileTree was changed in Pharo was changed to use STON directly for property files which results in the random ordering or the keys ...

fniephaus:
> at some point FileTree was changed in Pharo was changed to use STON directly for property files which results in the random ordering or the keys ...

@estebanlm any way to change this back? maybe this is a simple change?

PR #399, opened 2018-10-02
expose #ignoreImage metacello api

### LINKED ISSUE #333: Handling load conflicts in a SCIMetacelloLoadSpec (and other Metacello Scripting API functions)

estebanlm:
Fixes #333 (or trying ;) )

Please @dalehenrich and @fniephaus can you check at this?
I know there is a property mess (I need to re-fix that). But relevant part is at: 

```Smalltalk
SCIMetacelloLoadSpec>>#basicLoadProjectOn:
SCIMetacelloLoadSpec>>#ignoreImage
```


dalehenrich:
looks okay ... but I did make a comment on issue #333 ...

PR #411, opened 2019-02-02
Add Moose 7.0

badetitou:
Hi,

I've tried to change the files to add Moose 7.0
It's the first time i'm doing something like this, so it needs a real review.

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -81,7 +81,7 @@ moose::get_image_url() {
     "Moose32-trunk"|"Moose-trunk")
       moose_name="moose-7.0"
       ;;
-    "Moose32-6"*|"Moose-6"*|"Moose32-7"*|"Moose-7*")
+    "Moose32-6"*|"Moose-6"*|"Moose32-7"*|"Moose-7"*)
       moose_name="moose-$(echo "${smalltalk_name}" | cut -f2 -d-)"
       ;;
     *)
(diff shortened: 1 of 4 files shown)


fniephaus:
Hi @badetitou,
Thanks for the PR. Looks good so far. However, we are deprecating image names without architecture suffixes (for 32/64 bit), so we don't need to support Moose-7.0, only Moose32-7.0 and Moose64-7.0. BTW: are there 64bit Moose images?

badetitou:
There is a Moose 64 bit image. But I don't know if it's for public usage currently.
I would prefer to create a second pull request for Moose 64 later.

So, I suppose I should remove the image without architecture ?

pharo/run.sh
```diff
@@ -79,7 +79,10 @@ moose::get_image_url() {
 
   case "${smalltalk_name}" in
     "Moose32-trunk"|"Moose-trunk")
-      moose_name="moose-6.1"
+      moose_name="moose-7.0"
+      ;;
+    "Moose32-7"*|"Moose-7*")
```
fniephaus: s/"Moose-7*"/"Moose-7"* and this can be merged with the below case for Moose 6.

fniephaus:
Let's leave the Moose image without architecture for now. Sure, 64bit images can be added later.

badetitou:
I've fixed the code as you proposed @fniephaus

Issue #422, opened 2019-05-09
Coveralls results are not uploaded for Pharo64-7.0 and Pharo64-8.0


fjluquel:
Using Pharo64-7.0 or Pharo64-8.0 with Travis CI results in the following error message:

"Coveralls data is not available. Skipping coverage testing..."

The coverage data is not uploaded to Coveralls.io.

PR #423, opened 2019-05-09
Fix issue #422

fjluquel:
Use environment variable SMALLTALK_CI_BUILD for calculating absolute paths to find the coveralls files.


repository/SmalltalkCI-Coverage-Core.package/SCICodeCoverage.class/class/coverallsFilesDirectory.st
```diff
@@ -0,0 +1,3 @@
+coveralls
+coverallsFilesDirectory
+	^ (SmalltalkCI getEnv: 'SMALLTALK_CI_BUILD') asFileReference
```
gcotelli: `asFileReference` only works in Pharo, maybe you can reimplement this method in `SCIPharoCodeCoverage` class ?

fjluquel: Sure, I've already made the changes you suggested.

PR #433, opened 2019-07-30
Update Moose CI

badetitou:
Moose 32 8.0 doesn't exist
The url for Moose 8 does not follow the same format

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,6 @@ jobs:
     - smalltalk: Pharo32-3.0
     - smalltalk: Moose64-8.0
     - smalltalk: Moose64-7.0
-    - smalltalk: Moose32-8.0
     - smalltalk: Moose32-7.0
     - smalltalk: Moose32-6.0
     - smalltalk: GemStone64-3.4.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,6 +121,9 @@ jobs:
   allow_failures:
     - smalltalk: Squeak64-trunk
     - smalltalk: Pharo64-alpha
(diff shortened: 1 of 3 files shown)


fniephaus:
Some of the Moose tests did not pass. Not sure why, could you please have a look?

badetitou:
It looks like it's not moose the problem but Pharo 8.
The build of Pharo 8 is failing with the same error.

fniephaus:
The Pharo alpha builds fail due to `unknown option: --nodisplay` ([see here](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/565512226#L472)). Could someone familiar with the Pharo VM (@estebanlm?) comment on this? Why has it been removed and is there a replacement?

fniephaus:
@badetitou [This build](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/565512246#L667) failed due to a `download_file() expects an URL and a target path.` error. Could you please have a look at this?

badetitou:
@fniephaus we have decided to remove the build 32bits for Moose 8.0.
I forgot to remove the entry in .travis.yml
It should be fixed

badetitou:
It looks like it works when you ignore the coverage problem?

fniephaus:
> it looks like it works when you ignore the coverage problem?

Yes, but that's another problem which should be fixed at some point.
Anyway, thanks for the contribution!

fniephaus:
> it looks like it works when you ignore the coverage problem?

Yes, but that's another problem which should be fixed at some point.
Anyway, thanks for the contribution!

Issue #434, opened 2019-08-05
SCIPharoCodeCoverageTest>>testMethodReferenceForSelector fails on Pharo 8


fniephaus:
[Example build failure](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/565996310#L631).

Could someone from the Pharo community (maybe @jecisc) have a look at this please?

jecisc:
I tried to check the problem and apparently the Ring #name method changes of implementation in Pharo 8. 

If you want to get the selector of the method I think the best way is to use #selector instead of #name.

jecisc:
Reading the code of SmalltalkCI I have the impression that the real code is already using #selector instead of #name. I guess we can just update the test.

PR #435, opened 2019-08-05
434-SCIPharoCodeCoverageTesttestMethodReferenceForSelector-fails-on-Pharo-8

jecisc:
Update test to use the same methods than in the the code to test

Fixes https://github.com/hpi-swa/smalltalkCI/issues/434


fniephaus:
Thanks for the swift action, much appreciated! CI seems to pass AFAICT.

PR #453, opened 2020-04-21
Add support to upload coverage results to Coveralls if run as GitHub …

### LINKED ISSUE #452: Coveralls integration is not working when using GitHub actions to trigger the build
The API call to coveralls is not triggered because the build is not made by TravisCI nor Appveyor.

gcotelli:
…actions
Fixes #452

### DIFF
--- a/env_vars
+++ b/env_vars
@@ -3,7 +3,7 @@ if [[ -z "${SMALLTALK_CI_HOME:-}" ]]; then
 fi
 
 export SMALLTALK_CI_BUILD_BASE="${SMALLTALK_CI_HOME}/_builds"
-if [[ -z "${TRAVIS:-${APPVEYOR:-}}" ]]; then
+if [[ -z "${TRAVIS:-${APPVEYOR:-${GITHUB_ACTIONS:-}}}" ]]; then
   export SMALLTALK_CI_BUILD="${SMALLTALK_CI_BUILD_BASE}/$(date "+%Y_%m_%d_%H_%M_%S")"
 else
   export SMALLTALK_CI_BUILD="${SMALLTALK_CI_BUILD_BASE}"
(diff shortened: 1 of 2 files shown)


gcotelli:
Hi @fniephaus 
I hope you're fine in this world crisis.

I've managed to test this changes by making a release in my own fork, forking the github-action and using both in https://github.com/ba-st/Boardwalk/pull/20 . The smalltalkCI GitHub action was used to make the build and run the tests, and as a second step I've managed to grab the coverage output file and send it to CodeCov (see https://codecov.io/gh/ba-st/Boardwalk/tree/0a92b9568ff7472af3e9307f654f38704de9a405) . So now we have also the option to use this service instead of coveralls.

The only thing I didn't like right now is that if you have enabled the coverage it will always attempt to upload the resulting file to Coveralls, so if you're using another coverage service you can't skip this step. Maybe we can introduce an environment variable to disable this branch in the code?
I think it's better if the main smalltalkCI invocation just generates the coverage file (in case is enabled) and to be a second step to upload it to a coverage service provider but this will break backwards compatibility. What do you think?

fniephaus:
Hi @gcotelli,
Thanks for the PR and yes, all good. I hope you are well! Sorry for the delay, I will take a closer look this week.

Happy to add support for other coverage providers  if we can. Uploading coverage results is relatively simple and I think something smalltalkCI should help with if possible.  I'd like to avoid to support too many env variables, how about we add an `#service` option to the `#coverage` dict of `SmalltalkCISpec`? 

#### Example
```javascript
SmalltalkCISpec {
  ...
  #testing : {
    ...
    #coverage : {
      #classes : [ #ClassToCover, #'ClassToCover class' ],
      #service : #codecov
    }
  }
}
```

At the moment, supported services could be `#coveralls`, `#codecov`, and `#local` (for the coverage results file only). If we set the default to `#codecov`, we would even be backward compatible.

What do you think?

PR #455, opened 2020-04-27
Add support for additional coverage service providers

### LINKED ISSUE #454: Support for additional coverage services
@fniephaus moved from #453 to discuss it better: 
> Happy to add support for other coverage providers if we can. Uploading coverage results is relatively simple and I think something smalltalkCI should help with if possible. I'd like to avoid to support too many env variables, how about we add an #service option to the #coverage dict of  SmalltalkCISpec?
> Example
>
> SmalltalkCISpec {
>   ...
>   #testing : {
>     ...
>     #coverage : {
>       #classes : [ #ClassToCover, #'ClassToCover class' ],
>       #service : #codecov
>     }
>   }
> }
> 
> At the moment, supported services could be #coveralls, #codecov, and #local (for the coverage results file only). If we set the default to #coveralls, we would even be backward compatible.
> 
> What do you think?

I think there are 2 orthogonal issues here. One is the format of the exported coverage file (now it's always exported in coveralls JSON format) and I think that configuring it in SmalltalkCISpec will help with that if we want to support multiple export formats.
At least Codecov is capable of understanding the coveralls export format so we don't need to make changes per-se on the export code to support this service.
The second one is the service provider to which upload the exported coverage files. Now it is assumed that if you enabled coverage you want to upload it to coveralls and this is done in the shell scripts not in Smalltalk code. If I understand it right the info in SmalltalkCISpec is not available in the shell commands so I don't know how changing this spec will help in selecting the right provider to upload.

It always feels a bit strange to me that the smalltalkCI scripts per se are the ones uploading the files to the coverage services. For example for CodeCov if you're using GitHub actions you can re-use the official upload action and just provide the file name to upload in a later step. 

Maybe we can separate the upload code in a different shell script and add two command line options:
- One option will disable the automatic coverage upload inside the smalltalkCI script, so people reusing for example CodeCov GitHub actions can disable the automatic upload and let later steps take care of the upload.
- The other option will configure the coverage service provider, so if the automatic coverage upload is enabled this option is used to choose the service to call by invoking the chosen upload script.
- By default the automatic coverage upload will be `enabled` and the service provider will be `coveralls` to maintain backwards compatibility.

What do you think?

gcotelli:
Fixes #454


fniephaus:
Since you're testing against shellcheck, could we add a check to smalltakCI's CI? :)

gcotelli:
It was on the roadmap :) . I will see if the tests run and then add the GH action that run shellcheck against the scripts.

gcotelli:
@fniephaus I couldn't find the place where the info from the .ston config file is extracted in Bash. Did you have any hint? I will need it if I understood correctly.

gcotelli:
@fniephaus the PR is ready to review. Some jobs in Travis builds are failing but due to some issues downloading the VM :man_shrugging: 

The only thing I don't implement is "Add --upload-coverage hook to smalltalkCI to manually trigget the coverage file upload", because right know after the image ends running the tests it generates the script that it's used to perform the upload. So basically if the user wants to trigger manually the upload it can execute this script.

A brief description of the changes:
- Some changes are just the consequence of running Shellcheck (that now is integrated  as a GitHub action check)
- The `coverage` folder has the supporting Bash scripts. These scripts get copied to the build directory during the run, because `upload-coverage.sh` is a placeholder that is modified with Smalltalk code. 
- `upload-to-coveralls.sh` contains the code previously used to make the upload
- `upload-to-codecov.sh` contains the recommended Bash uploader code (https://docs.codecov.io/docs/about-the-codecov-bash-uploader)
- The smalltalk changes are pretty straightforward: Support for the new coverage options and generating `upload-coverage.sh` contents by sourcing the right script.

fniephaus:
Cool, this is great! I will have a closer look this week. :)

gcotelli:
I've tested it using the new capabilities we added to the GitHub action to use this branch and an ad-hoc repository with smalltalk code. So far it seems to be working given you have the proper credentials configured.

fniephaus:
Well done! I promise to have a look this week :)

README.md
```diff
@@ -515,6 +515,25 @@ SmalltalkCISpec {
...
+    #coverage : {
+      #packages : [ 'Packages-To-Cover.*' ],
+      #classes : [ #ClassToCover, #'ClassToCover class' ],
+      #categories : [ 'Categories-To-Cover*' ],
+      #service: #codecov
+    }
+  }
+}
+```
+If the configured service is not `#local`, the `#auto_upload` key will control if the coverage report will be uploaded automatically as part of the test run step or no. In practical terms setting `#auto_upload` to false is equivalent to use `#local` as service.
```
fniephaus, line 536: > step or no
I don't understand this, is something missing here?

bin/smalltalkci
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [[ -z "${SMALLTALK_CI_HOME:-}" || "$@" = *--self-test* ]]; then
+if [[ -z "${SMALLTALK_CI_HOME:-}" || "$*" = *--self-test* ]]; then
```
fniephaus, line 3: Does this fix an actual problem?

coverage/coverage-utils.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
fniephaus, line 1: Couldn't we just re-use `helpers.sh` instead of this file?

coverage/upload-to-coveralls.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+# shellcheck source=coverage-utils.sh
+source "${SMALLTALK_CI_BUILD}/coverage-utils.sh"
+
+readonly COVERALLS_API="https://coveralls.io/api/v1/jobs"
+readonly coverage_results="${SMALLTALK_CI_BUILD}/coveralls_results.json"
```
fniephaus, line 7: `coverage_results` definitions are inconsistent, sometimes with `readonly` sometimes not. Let's use `readonly COVERAGE_RESULTS`, it's a constant, right?

helpers.sh
```diff
@@ -353,7 +364,16 @@ git_log() {
 }
 
 
-export_coveralls_data() {
+prepare_coverage_data() {
+
+  print_info "Preparing coverage templates..."
+  cp "${SMALLTALK_CI_HOME}"/coverage/* "${SMALLTALK_CI_BUILD}"
+  chmod +xw "${SMALLTALK_CI_BUILD}/upload-coverage.sh"
```
fniephaus, line 371: The `chmod`s should not be necessary if they are executable (and writable in the first case) on your system. The permissions should be stored correctly in Git.

repository/SmalltalkCI-Coverage-Core.package/SCICodeCoverage.class/instance/coverageReportUploadActionScript.st
```diff
@@ -0,0 +1,11 @@
...
+coverageReportUploadActionScript
+
+	"This script is sourced in 'upload-coverage.sh script and executed after the image is closed."
+	self uploadsCoverageReportAutomatically
+		ifTrue: [ 
+			self serviceName = #coveralls
+				ifTrue: [ ^ 'upload-to-coveralls.sh' ].
+			self serviceName = #codecov
+				ifTrue: [ ^ 'upload-to-codecov.sh' ] ].
+	^ 'skip-coverage-upload.sh'
```
fniephaus, line 11: Mh, now that I see this, I don't like the fact that the Smalltalk knows about bash script names. Could we rename `upload-coverage.sh` to something like `coverage.env` and store `readonly COVERAGE_SERVICE="{serviceName}"` in there? On the bash side, we can then source `coverage.env` and check `$COVERAGE_SERVICE` for its contents. What do you think?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Shellcheck'
+
+on: [push,pull_request]
+
+jobs:
+  shellcheck:
```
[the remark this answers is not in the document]
gcotelli: As you wish. This way it runs in parallel and the feedback is more immediate because if it fails you don't have to dig in the logs to discover that was the shellcheck.

README.md
```diff
@@ -515,6 +515,25 @@ SmalltalkCISpec {
   }
 }
 ```
+The coverage service will default to [coveralls.io][coveralls] to keep backwards compatibility. But you can use:
+- `#coveralls` will use [coveralls.io][coveralls]
+- `#codecov` will use [codecov.io][codecov]
+- `#local` will not attempt to upload the coverage report to any service. Useful for local runs or to setup the upload as a separate step in the build.
```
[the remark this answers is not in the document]
gcotelli: I think it's a good debugging feature, and without setting the environment variables with the right tokens it will fail anyway if you run it locally by mistake.

README.md
gcotelli: I think that probably I want to put "or not" ?

bin/smalltalkci
gcotelli: It was reported by shellcheck:  `SC2199: Arrays implicitly concatenate in [[ ]]. Use a loop (or explicit * instead of @).`

helpers.sh
gcotelli: That was my intuition but somehow the permissions got lost when running in the GitHub action, but can be just that I've run this several times until I got it working. Debugging things running on a remote server without any kind of interaction possible sucks :( . If you want to remove it, do it in an isolated commit so it can be easily recovered if don't work.

repository/SmalltalkCI-Coverage-Core.package/SCICodeCoverage.class/instance/coverageReportUploadActionScript.st
gcotelli: I think it's a good idea. It was the first thing I've proposed: using some env variables. I have some busy weeks ahead so if you want to proceed with the proposed changes go ahead.

coverage/coverage-utils.sh
gcotelli: Now that everything is working we can probably use helpers. I was trying to not break any other thing while I was testing the changes. The only difference in the functions is in the detection of terminals without color support.

MrModder:
After merging #466, it is now possible to upload code to CodeCov and other  service providers using the LCOV format. And whilst you have to upload the LCOV data manually, it is pretty simple (as documented in [COVERAGE.md](https://github.com/hpi-swa/smalltalkCI/blob/master/docs/COVERAGE.md) ).
So I think this PR can be closed. We don't really want to add code to upload to all providers from smalltalkCI directly, as this would be a lot to maintain.

@gcotelli do you agree?

fniephaus:
Closing this for now...we can always reopen it or create a new PR if there's something that should be integrated into smalltalkCI.

Issue #464, opened 2020-08-26
Introduce multiple Coverage output formats


MrModder:
In order to increase compatibility for the possible combinations of CI infrastructure+Coverage Reporters, smalltalkCI is aiming to support multiple Coverage output formats (current targets: Coveralls + LCOV).

In future we will no longer aim to upload all these formats to all different Coverage reporters, as many uploaders already exist for this purpose.
We should instead focus on providing good documentation on how to use smalltalkCI with this existing infrastructure.
The current capabilities to upload to Coveralls, when using TravisCI will remain for backwards compatibility.
This would therefore also close #454 .

Steps to achieve this goal:
- Add abstraction to write multiple formats (PR: #463)
  - [x] Remove export functionality from SCICodeCoverage
  - [x] Introduce abstract SCICoverageWriter
  - [x] Introduce SCICoverallsWriter
- Add LCOV format writer
  - [x] Introduce SCILcovWriter
  - [x] Allow user to choose coverage output format in `.smalltalk.ston`
  - [x] Document accordingly

LCOV Coverage should also close #458, as the uploaders for both [Coveralls](https://github.com/marketplace/actions/coveralls-github-action) and [CodeCov](https://github.com/marketplace/actions/codecov) can be used.

PR #466, opened 2020-08-28
Introduce LCOV coverage writer

MrModder:
This PR aims to implement the second steps of issue #464 
- [x] Introduce SCILcovWriter
- [x] Allow user to choose coverage output format in `.smalltalk.ston`
- [x] Document accordingly


COVERAGE.md
```diff
@@ -0,0 +1,182 @@
...
+```
+If you have LCOV installed, you can then generate the report:
+```bash
+cd /path/to
+genhtml lcov.info
+xdg-open index.html
+```
+The result will look something like this:
+
+<img src="resources/lcov-example.png" alt="An example LCOV report" width=75% />
```
fniephaus, line 68: Could we please move the `lcov-example.png` to https://github.com/hpi-swa/smalltalkCI/tree/assets? It does not need to be part of the master branch, which is pulled every time a project is tested

MrModder: `lcov-example.png` has been moved to assets issue #467

fniephaus (approved):
LGTM, thanks! Will push some minor cleanups after merging...

PR #469, opened 2020-09-03
Add instructional prompt after loading project

### LINKED ISSUE #456: --headful walkback if image is not saved and missing docs
It took some time until I understood, that, after loading my code with the --headful parameter, I have to **save and quit** the image for starting the tests.

I was starting smalltalkCI with 

`/bin/smalltalkci --headful --debug  --smalltalk "Pharo64-7.0" /../.smalltalk.ston`

and after loading, the image does nothing. I would expect that there is a dialog or at least a sentence in the documentation explaining that I have to "save and quit" for it to proceed with the tests.

I was trying to leave the Image without saving, then I got a walkback similar to 
`https://github.com/pharo-project/pharo/issues/2720`

Stack:
[ (FileWriteError fileName: self name)
	signal:
		(self closed
			ifTrue: [ 'File ' , self name , ' is closed' ]
			ifFalse: [ 'File ' , self name , ' write failed' ]) ] in StdioStream(AbstractBinaryFileStream)>>next:putAll: in Block: [ (FileWriteError fileName: self name)...
BlockClosure>>cull:
Context>>evaluateSignal:
Context>>handleSignal:
PrimitiveFailed(Exception)>>signal
PrimitiveFailed class(SelectorException class)>>signalFor:
File class(ProtoObject)>>primitiveFailed:
File class(ProtoObject)>>primitiveFailed
File class>>write:from:startingAt:count:
[ File
	write: handle
	from: aByteArray
	startingAt: 1
	count: amount ] in StdioStream(AbstractBinaryFileStream)>>next:putAll: in Block: [ File...
BlockClosure>>on:do:
StdioStream(AbstractBinaryFileStream)>>next:putAll:
StdioStream(AbstractBinaryFileStream)>>nextPutAll:
StdioStream(AbstractBinaryFileStream)>>nextPut:
ZnUTF8Encoder>>next:putAllASCII:startingAt:toStream:
ZnUTF8Encoder>>next:putAllByteString:startingAt:toStream:
ZnUTF8Encoder>>next:putAll:startingAt:toStream:
ZnCharacterWriteStream>>next:putAll:startingAt:
ZnCharacterWriteStream(ZnEncodedWriteStream)>>nextPutAll:
ZnNewLineWriterStream>>newLine
ZnNewLineWriterStream>>nextPut:
[ :v | self nextPut: v ] in ZnNewLineWriterStream(Stream)>>nextPutAll: in Block: [ :v | self nextPut: v ]
ByteString(SequenceableCollection)>>do:
ZnNewLineWriterStream(Stream)>>nextPutAll:
ZnNewLineWriterStream(WriteStream)>>nextPutAll:
SmalltalkCIPharo78 class(SmalltalkCI class)>>fold:on:block:
SmalltalkCIPharo78 class(SmalltalkCI class)>>fold:block:
SmalltalkCIPharo78 class(SmalltalkCI class)>>stage:id:block:
SmalltalkCIPharo78(SmalltalkCI)>>basicLoad
[ super basicLoad ] in SmalltalkCIPharo78(SmalltalkCIPharo)>>basicLoad in Block: [ super basicLoad ]

LeonMatthes:
Closes #456


LinqLover:
Nice idea! Would it be helpful to print a message to stdout, too? When I'm running something from the terminal, I'll often expect the terminal to print any message when I'm required to take some action.

MrModder:
> Nice idea! Would it be helpful to print a message to stdout, too? When I'm running something from the terminal, I'll often expect the terminal to print any message when I'm required to take some action.

I don't really see the reason for doing this.
If you're explicitly running in headful mode, I'd expect you to be able to see and pay attention to the running image.
Why display it on stdout as well?

LinqLover:
> > Nice idea! Would it be helpful to print a message to stdout, too? When I'm running something from the terminal, I'll often expect the terminal to print any message when I'm required to take some action.
> 
> I don't really see the reason for doing this.
> If you're explicitly running in headful mode, I'd expect you to be able to see and pay attention to the running image.
> Why display it on stdout as well?

I was thinking of several modifications that can occur during test execution and break any part of the UI. E.g. a test could install a different ToolSet and forget to reset it or modify the visual appearance in some way that makes the UIManager prompt invisible. In this case, debugging would be easier if a hint was printed into the terminal, too.

Another reason is that the test execution can take a long time and while you are working in another program, the VM usually does not inform you when a dialog window appears. In contrast, many terminals can notify the user if a new portion of output was printed to a background terminal instance (for example, https://github.com/felixse/FluentTerminal/issues/60 or https://github.com/microsoft/terminal/issues/7955).

A third use case would be UI tests that disallow you to interact with the image which could have side effects on the tests. In such situations, I use to minimize the VM window while the tests are executing, so I would not see the VM prompt, again, but only any terminal message.

Do you see what I mean? :-)

LinqLover:
Hm, I just realized that the same claim could also be made for all the progress bars that appear during the image setup/project loading stage. We could also duplicate all kinds of outputs in an extra ToolSet implementation and avoid any extra logic in this PR. In any case, it was not my intention to block this PR. :-)

LeonMatthes:
Okay, you bring up some fair points. Most of them focus on the `testing` stage, however this PR does not change this.
All UI elements in this PR open after the `installation` and before the `testing` stage.

There actually is (or was at some point) a UIManager that uses the Command-line instead of Squeak GUI elements, so that one could be installed as well.
It's most likely best to try that route first, as it would also show other dialogs on the command line, and maybe even progress bars (not sure about that though).

LinqLover:
> All UI elements in this PR open after the installation and before the testing stage.

Ah, I missed that, fair point. There should be a rare number of installation scripts that corrupt the overall UI appearance.

> There actually is (or was at some point) a UIManager that uses the Command-line instead of Squeak GUI elements, so that one could be installed as well.

Afaik there is no such implementation for Squeak, but one for Pharo. But a `CommandLineUIManager` is already on my wish-list, see <http://forum.world.st/CommandLineUIManager-td5106538.html>. I would be glad to both implement it for Squeak and integrate it in SCI. :-)

tl;dr: Please pardon the interruption and go ahead with your changes! :-)

fniephaus (approved):
LGTM, thanks a lot for this!

I've changed `SmalltalkCI>>#promptToProceedImpl` so that proceeds by default, reverted redundant changes to various metadata files, and fixed the author initials.

I'll merge this once CI has passed again...

PR #484, opened 2020-11-02
fix system reporter for Pharo 9

demarey:
SystemReporter class was just removed from Pharo 9. Let's fix it.


fniephaus:
I've pushed a commit to restore and fix the metadata. Will merge this when CI passes...

fniephaus (approved):
LGTM

demarey:
Thanks Fabio. A temporary fix was pushed into Pharo 9.
A more permanent fix will be done into Pharo but.it will have no impact on smalltalk ci.
We should forget this PR for now.

fniephaus:
What's the "more permanent fix"? We ran into a similar issue with `TestCase`...

demarey:
The right fix will be to split the tool in 2 packages: one for the UI, one for the business code.
See https://github.com/pharo-project/pharo/pull/7659

fniephaus:
Excellent, and thanks for the reference!

PR #485, opened 2020-11-07
SCIMetacelloLoadSpec: Add parameters for #username and #password

LinqLover:
This PR adds to new parameters to `SCIMetacelloLoadSpec`, `#username` and `#password`. Example usage in a `.smalltalk.ston` file:

```smalltalk
SCIMetacelloLoadSpec {
	#baseline : 'YourRepository',
	#directory : 'src',
	#platforms : [ #squeak ],
	#load : [ #tests ],
	#user : 'yourfancyusername',
	#password : 'yourfancypassword'
}
```

Complements and depends on Metacello/metacello#536.


LinqLover:
I did not find any tests for the actual loading logic, i.e. `SCIPlatformDependent >> #loadProjectOn:` and `#basicLoadProjectOn:`. Should such tests exist? If yes, should they be unit tests (and set up a quite large mocking environment) or acceptance tests (which would require the creation of example repositories on several platforms such as GitHub or a Monticello server)? Or aren't such tests necessary/profitable at all? :-)

fniephaus:
Looks like this requires the latest Metacello? We could either ensure the latest Metacello is used when username/password are provided or check for support via respondsTo:? What do you suggest?

gcotelli:
I think it's not a good idea to put a password in plain text in a config
file.

fniephaus:
I have to agree with @gcotelli on this. This encourages users to store credentials in a config file. CI providers often support secrets either through env vars or files. Maybe Metacello needs to provide support for one of these approaches?

LinqLover:
> I have to agree with @gcotelli on this. This encourages users to store credentials in a config file. CI providers often support secrets either through env vars or files. Maybe Metacello needs to provide support for one of these approaches?

Yep, I have already implemented this in my working copy with env vars, too. Looking forward to submit my changes as soon as possible in a second PR! :D

LinqLover:
> Looks like this requires the latest Metacello? We could either ensure the latest Metacello is used when username/password are provided or check for support via respondsTo:? What do you suggest?

Wouldn't it be possible to update the prepared images with the latest Metacello version instead? Otherwise, I can add a hook to `updateMetacelloIfRequested`, but I think this would be a workaround as soon as the smalltalkCI images will have been updated somewhen in the future. :-)

fniephaus:
Updating all images is possible, however we've tried to avoid that if possible in the past. Note that smalltalkCI loads Pharo images from Pharo servers, and they may not be updated accordingly. I'd prefer a consistent solution.

LinqLover:
> > I have to agree with @gcotelli on this. This encourages users to store credentials in a config file. CI providers often support secrets either through env vars or files. Maybe Metacello needs to provide support for one of these approaches?
> 
> Yep, I have already implemented this in my working copy with env vars, too. Looking forward to submit my changes as soon as possible in a second PR! :D

Please see #486. :-)

LinqLover:
@fniephaus Metacello/metacello#536 has been merged and now the CI passes! 🎉 Looking forward to your review or merge! :-)

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/password.st
```diff
@@ -0,0 +1,4 @@
+accessing
+password
+
+	^ password ifNil: [ ^ '' ]
```
fniephaus, line 4: Why return an empty string if nil? Wouldn't it make all checks easier if it would return `nil` when the password is not set?

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/requiresLatestMetacello.st
```diff
@@ -0,0 +1,11 @@
+private
+requiresLatestMetacello
+	"Check for missing features."
+
+	"Credentials parameters (https://github.com/Metacello/metacello/pull/536)"
```
fniephaus, line 5: I wonder: should smalltalkCI print something to stdout informing the user why it has decided to upgrade Metacello? Something like:
> Upgrading to the latest Metacello because credentials are provided...

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/user.st
```diff
@@ -0,0 +1,4 @@
+accessing
+user
+
+	^ user ifNil: [ ^ '' ]
```
fniephaus, line 4: Same here: wouldn't it make things easier to return `user` directly?

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/basicLoadProjectOn..st
```diff
@@ -12,17 +12,21 @@ basicLoadProjectOn: aSmalltalkCI
 		baseline: self baseline;
 		configuration: self configuration;
 		version: self version.
-
+	self user ifNotEmpty: [:currentUser |
```
fniephaus, line 15: Does it make sense to only provide `user` and not or `password` (or vice-versa)? I mean can you access a private repo by only providing a `password` or `user`? I'd be surprised if that works for `user`, but it may work for `password` if it's set to some API token. If some combination is invalid (e.g. when only a `user` is provided), is Metacello failing the smalltalkCI build with a good error message?

fniephaus (changes requested):
Excellent PR, thanks a lot @LinqLover! Only one question and some minor remarks.

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/password.st
LinqLover: This implementation was copied from `SCIMonticelloLoadSpec` where you fell back to `''` as well. :-) But yes, I would prefer to return `nil`, too, I will change that.

LinqLover: (I don't know whether Monticello or Metacello depend on a non-nil string, but to prevent this, we do not need to pollute the public accessor protocol of SCI.)

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/requiresLatestMetacello.st
LinqLover: Good idea! I will implement this soon.

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/password.st
fniephaus: Good catch, maybe that should be cleaned up at some point. It looks like `MCHttpRepository` can handle `nil` values for `#password` and `#user`. If you like, you can clean this up as part of this PR.

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/basicLoadProjectOn..st
LinqLover: > but it may work for password if it's set to some API token.

Yep, I'm using this a lot because it saves me the time to define another CI secret for my username. See https://github.com/Metacello/metacello/blob/master/repository/Metacello-Platform.squeak.package/MCFilesystemFetchOnlyRepository.class/class/sitePassword.st. :-)  
For username only, I'm not sure if there any scenarios, but I would rather not forbid this without a good reason (and if at all, this should be rather implemented in Metacello, shouldn't it?).

> is Metacello failing the smalltalkCI build with a good error message?

It depends. For an invalid password, for example, it says:

```
gofer repository error: 'GoferRepositoryError: Error downloading xxx to yyy.zip. Server said: {"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/repos#download-a-repository-archive"}'...ignoring
```

In this case, the repository was private, so the API does not distinguish between inaccessible and inexistent for privacy reasons, but otherwise, the message would be something like "Not authorized".

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/password.st
LinqLover: I could do this, however, #486 touches exactly the same methods, so I'd rather refactor the nil values in that PR ...

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/basicLoadProjectOn..st
fniephaus: Ok, it should really be Metacello's job to validate the input anyway :)

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/password.st
fniephaus: Right!

repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/user.st
LinqLover: IIRC for some client is was important not to pass `nil` as a user. Nevertheless, I will take a second look at this when implementing `SCIAbstractLoadSpecWithCredentials` for #486 when I find some time for it :-)

LinqLover:
This PR depends on #486 and already has the changes from the other branches merged. Will be ready for review as soon as #486 got merged.

Please squash when merging. Working with FileTree repositories can be so hard, everything needs to be committed + merged twice ... :-)

fniephaus (approved):
LGTM, thanks!

PR #496, opened 2020-12-27
Fix handling of CTRL + C during smalltalk image prompt

LinqLover:
For more information on this bug see: https://unix.stackexchange.com/a/513498/448260


fniephaus (approved):
LGTM, thanks for the fix!

PR #497, opened 2021-01-02
squeak: add basic support for tonel

tom95:
If a repo that advertises itself as being in the tonel format via the .properties file is encountered, we load the tonel library and adapt the project spec string accordingly.

Note that I have not tested coverage or anything else.


tom95:
Note that I had to adapt a test to change a path to use `\` when running on Windows machines, which was not necessary before as `#filetreeRepositoryIn:` used to not access the filesystem.

fniephaus (approved):
LGTM, thanks!

fniephaus:
It appears that this has broken GemStone builds. I am working on a fix...

Issue #471, opened 2020-09-24
Cache prepared testing images for Squeak-Trunk?


LinqLover:
Has this ever been discussed in the past? For builds based on a Squeak-Trunk image, every single CI build takes nearly 60 seconds just for "preparing [the image] for CI" (see `./squeak/prepare.st`).

Wouldn't it be an option to cache these images after created once? I could imagine two different approaches to do this:

1. Automatically release new prepared images in this repository on a high-frequent basis (e.g., once per hour), maybe on a separate branch. Refer to their location in [`run_build()`](https://github.com/hpi-swa/smalltalkCI/blob/f343d8a8b315e4a1184668e4a3ce34b92ad95332/squeak/run.sh#L351).

   Pros:
    - No external architecture needed.

   Cons:
    - I don't know this would interfere with the current release process.
    - To what extent is the release process automated at the moment? For example, did you create the `.tar.gz` files in https://github.com/hpi-swa/smalltalkCI/releases/tag/v2.9.4 manually, or is this hidden somewhere in the `.travis.yml` file and I did not find it?
     - Not yet handled Trunk updates still would trigger manual image preparation for every single build. I don't know how many builds this actually affects, since you do not list this information on https://smalltalkci.fniephaus.com/.

2. Outsource the preparation of trunk images to some kind of external server that uses a cache, i.e. only prepares every Trunk revision once.

   Pros:
    - Best optimized solution

   Cons:
    - Custom server needed -> configuration & maintenance effort
    - Communication to that server needs to be implemented

By the way, can you specify an arbitrary Trunk revision number for smalltalkCI at the moment, without downloading the image manually, something like `smalltalk: Squeak64#19851`? I believe this could be helpful, too. (For example, in https://github.com/LinqLover/Squeak-TipOfTheDay, I require #18899 or higher, so I would test both Trunk and #18899 to test backward compatibility).

PS: If a useable solution arises from this, it could possibly also be reused for other Smalltalk dialects.

fniephaus:
Hi @LinqLover,

From time to time, we bump the trunk image for smalltalkCI to reduce the time to update to the latest trunk. I don't think caching is a good idea, it often only leads to new problems.

[...]

PR #506, opened 2021-03-25
Squeak: Specify exact Trunk image version

LinqLover:
Trying to realize the idea I mentioned here: https://github.com/hpi-swa/smalltalkCI/issues/471#issuecomment-699024288

Usage: `smalltalkci -s "Squeak{{ 32 | 64}}-{{ version_number }}`, where `version_number` points to a Squeak version such as `19431`.

Work in progress, let's see whether it works ... :-)

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -193,7 +193,27 @@ is_sudo_enabled() {
   $(sudo -n true > /dev/null 2>&1)
 }
 
+is_tagged_build() {
+  if [[ ([[:digit:]]+) =~ "${config_smalltalk}" ]]; then
+    export SCIII_SMALLTALK_VERSION="${BASH_REMATCH[1]}"
+    return 0
+  fi
+  return 1
+}
+
 is_trunk_build() {
+  if is_tagged_build; then
+    case "${config_smalltalk}" in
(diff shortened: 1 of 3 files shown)


LinqLover:
Only one failure left which seems unrelated to my changes. Apart from that, the `MCMcmUpdater>>#doUpdate:upToUpdate:` proposal is pending discussion on squeak-dev: [The Inbox: MonticelloConfigurations-ct.166](http://forum.world.st/The-Inbox-MonticelloConfigurations-ct-166-mcz-td5128353.html)

fniephaus:
What happens if I request a trunk image with update level less than the one of the base image that smalltalkCI uses? We'll probably have to update trunk base images some time in the future again, so this is quite likely. I don't think we can downgrade images, or can we?

LinqLover:
Afaik downgrading is not possible. My current approach always uses the latest image that was released before the requested version number, see `squeak::latest_release_before()` (https://github.com/hpi-swa/smalltalkCI/pull/506/files#diff-f15a119f2269fdd82da32c9e83444165027e5d69e93ecc94bed8da5f86761ade) :-)

PR #507, opened 2021-03-31
Update Moose8 Stable location

badetitou:
I have updated the Moose8 stable continuous release location

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -100,7 +100,7 @@ moose::get_image_url() {
         echo "https://ci.inria.fr/moose/job/${moose_name}/lastSuccessfulBuild/artifact/${moose_name}.zip"
       ;;
      "Moose64-8"*)
-      echo "https://ci.inria.fr/moose/view/Moose%208.0/job/Moose-8/ARCHITECTURE=64,PHARO=80,VERSION=development,VM=vm/lastSuccessfulBuild/artifact/Moose.zip"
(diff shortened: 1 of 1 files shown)


badetitou:
I do not know why there is a conflict problem with the last Moose64-8
Maybe because the image is build using smalltalkCI? 

Any idea

fniephaus:
It looks like the [new image](https://github.com/moosetechnology/Moose/releases/download/v8.x.x/Moose8-stable.zip) has some version of smalltalkCI pre-loaded. Why's that?

badetitou:
Indeed. The image is built using the smalltalk ci project. (we use smalltalk ci to create the image )
So the project is loaded

fniephaus:
Yes, smalltalkCI needs to replace itself. I will take a look...

badetitou:
Super!

It looks like a great feature to me if we can build image using smalltalkci 
I'll try to give you a hand if I found time beside my manuscript thesis writing

fniephaus:
Merged via https://github.com/hpi-swa/smalltalkCI/commit/69ad6ca1ebe6de4740b074032a3cfc875e96e0a5. Looks like you forgot about Moose64-trunk ;)

badetitou:
Thanks for the update, We are moving to a new version of Moose (Moose9 based on Pharo9), so the trunk will evolve. But we will see this in a future PR. 
Thanks !

PR #508, opened 2021-04-04
TestSuite class should not be hard coded

mattonem:
We should be able to run tests using a custom test suite class, that defines setup and teardown, run test in parallel, etc...

In my case, I'm doing e2e testing with Parasol (and Browserstack) of my Seaside webserver. My test cases could be run in parallel most of the time. The overall pipeline time could improve by a tone. And I could increase device coverage without increasing the time I spend executing in the pipeline.

I'm not sure about the contribution process, I tried using Pharo&Iceberg but the diff was all messed up... maybe due to tonel format...
So I had to redo the contribution using Github online edit... I'm not sure it works... But the idea is here. Please let me know about the contribution process, I'd be glad to redo the work.


mattonem:
#509 might be a better implementation option for supporting custom test suite as it relies on the testcase ability of self-building its own test suite

PR #509, opened 2021-04-04
TestCase might require custom suite to be run

mattonem:
Another workaround for supporting custom Test suites

### DIFF
--- a/repository/SmalltalkCI-Core.package/SCITestRunner.class/class/runClasses.spec..st
+++ b/repository/SmalltalkCI-Core.package/SCITestRunner.class/class/runClasses.spec..st
@@ -6,5 +6,5 @@ runClasses: aCollectionOfClasses spec: aSCISpec
     select: [ :each | (each includesBehavior: TestCase) and: [ each isAbstract not ] ])
       asSortedCollection: [ :a :b | a name <= b name ].
   classes
-    do: [ :each | each addToSuiteFromSelectors: suite ].
+    do: [ :each | suite addTests: each buildSuite ].
   ^ self runSuite: suite spec: aSCISpec
--- a/repository/SmalltalkCI-Core.package/SCITestRunner.class/class/runClasses.spec..st
+++ b/repository/SmalltalkCI-Core.package/SCITestRunner.class/class/runClasses.spec..st
@@ -6,5 +6,5 @@ runClasses: aCollectionOfClasses spec: aSCISpec
     select: [ :each | (each includesBehavior: TestCase) and: [ each isAbstract not ] ])
       asSortedCollection: [ :a :b | a name <= b name ].
   classes
-    do: [ :each | suite addTests: each buildSuite ].
+    do: [ :each | suite addTest: each buildSuite ].
   ^ self runSuite: suite spec: aSCISpec
--- a/repository/SmalltalkCI-Core.package/SCITestReporterXML.class/instance/printStart.st
+++ b/repository/SmalltalkCI-Core.package/SCITestReporterXML.class/instance/printStart.st
@@ -3,7 +3,7 @@ printStart
 	self
 		print: '<?xml version="1.0" encoding="UTF-8"?>'; newLine;
 		print: '<testsuite name="', (self encode: runner suiteName);
-    		print: '" tests="', runner suite tests size asString;
+    		print: '" tests="', runner totalTests asString;
     		print: '" failures="', runner failedTests asString;
     		print: '" errors="', runner erroredTests asString;
     		print: '" time="', (runner suiteTime / 1000.0) asString , '">';
--- a/repository/SmalltalkCI-Core.package/SCITestRunner.class/instance/totalTests.st
+++ b/repository/SmalltalkCI-Core.package/SCITestRunner.class/instance/totalTests.st
@@ -1,3 +1,7 @@
 accessing
 totalTests
-	^ self suite tests size
+	| size |
+	size := 0.
+	results valuesDo: [ :clsResults |
+		size := size + clsResults size ].
+	^ size


mattonem:
It should be perfectly fine for a test suite to contain other instances of test suites:

> That is, a test suite contains sub-instances of TestCase and TestSuite. Both individual TestCases and TestSuites understand the same protocol, so they can be treated in the same way; for example, both can be run. This is in fact an application of the composite pattern in which TestSuite is the composite and the TestCases are the leaves — see Design Patterns for more information on this pattern2 .
_[Sunit doc](http://pharo.gforge.inria.fr/PBE1/PBE1ch8.html)_

fniephaus:
Hi @mattonem,

Thanks for the PRs and sorry for the slow response. I'm assuming that with this, smalltalkCI now works for your use cases?

Best,
Fabio

fniephaus (approved):
LGTM. Can this be merged?

repository/SmalltalkCI-Core.package/SCITestRunner.class/class/runClasses.spec..st
```diff
@@ -6,5 +6,5 @@ runClasses: aCollectionOfClasses spec: aSCISpec
     select: [ :each | (each includesBehavior: TestCase) and: [ each isAbstract not ] ])
       asSortedCollection: [ :a :b | a name <= b name ].
   classes
-    do: [ :each | each addToSuiteFromSelectors: suite ].
-  ^ self runSuite: suite spec: aSCISpec
\ No newline at end of file
+    do: [ :each | suite addTest: each buildSuite ].
```
fniephaus, line 9: Hang on... why does this work? This adds a suite to a collection of tests, right?

fniephaus (changes requested):

mattonem:
> Hi @mattonem,
> 
> Thanks for the PRs and sorry for the slow response. I'm assuming that with this, smalltalkCI now works for your use cases?
> 
> Best,
> Fabio

Hey, 
Thanks a lot @fniephaus,
I must be honest, I'd like to have it thoroughly tested before merging.
I couldn't find an easy way to execute a local SmalltalkCI against a local smalltalk project... 
I think I will try to add a CustomTestSuite to SmalltalkCI tests. 
I'll update the PR asap.

mattonem:
Hi @fniephaus,
It took ages to figure that one out. It looks like Squeak test suite is very different from Pharo test suite...
- For a first, squeak test suite does not support setUp and tearDown -- yet it is possible to achieve something equivalent, overwriting the run method. 
- Secondly, the squeak default will do the same thing as SmalltalkCIRunner currently (create its own standard test suite and inject test case by case)... Pharo, on the other hand, would take custom testSuite into account.

Those observation make the contribution less valuable for sure... it would still work, but the test runner of squeak and the test runner of smalltalkCI would have different behavior... 

I guess I could make the contribution in the Pharo package, that would make more sense. Trying to align the behavior smalltalkCi with the behavior of the default language test runner... 

Any thoughts on that?

fniephaus:
> It looks like Squeak test suite is very different from Pharo test suite...

I'm afraid I was expecting that. However, the class comment of Squeak's `TestSuite` mentions:
*... The common protocol is #run: aTestResult and the dependencies protocol*

and `#run:` calls `setUp` and `tearDown` correctly.

> the squeak default test runner (the graphical one) will do the same thing as SmalltalkCIRunner currently (create its own standard test suite and inject test case by case)... Pharo, on the other hand, would take custom testSuite into account.

This indicates that we might need dialect-specific runners in smalltalkCI. I guess calling `testCase buildSuite` and then adding all tests from that suite to smalltalkCI's suite defeats the purpose and won't solve your problem.

Anyway, do we have any more examples for custom test suites? It may be useful to know how they are used in the wild.

LinqLover:
> Anyway, do we have any more examples for custom test suites? It may be useful to know how they are used in the wild.

I have used [custom test suites](https://github.com/Museum-Barberini-gGmbH/Barberini-Analytics/blob/1cb0bbdd18bea4b7096885f4bf32a648ee114291/tests/_utils/db_test.py#L121) in the past for providing suite-wide isolation mechanisms / setUps & tearDowns. (To be fair, that was Python/unittest, but the basic architecture is still very similar to the original SUnit architecture.)

mattonem:
Well, Parallel testing is definitively something that can only be achieved with custom TestSuite as it requires to run testCases in different manner. 
But another scenario I encountered, is dealing with remote complex systems (like a robot made of multiple actuator).
Ideally, to have:
- first TestCase named `RobotStartAndStopTest` which only test start and stop procedure.
- then another TestCase named `RobotActionTest` which would start the robot at beginning of the test suite (a set up) and execute a serie of functional tests and stop the robot (test suite tear down.)

In Smalltalk world the second test case would ideally be implemented using TestResource and not CustomTestSuite.
It would **_ideally_** work like this:
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVGVzdFN1aXRlXG4gICAgcGFydGljaXBhbnQgVGVzdENhc2VTdGFydEFuZFN0b3BcbiAgICBwYXJ0aWNpcGFudCBUZXN0Q2FzZUFjdGlvbnNcbiAgICBwYXJ0aWNpcGFudCBUZXN0UmVzb3VyY2VcbiAgICBwYXJ0aWNpcGFudCBSb2JvdFxuXG4gICAgVGVzdFN1aXRlIC0-PiArVGVzdENhc2VTdGFydEFuZFN0b3A6IHJ1blxuICAgIGFjdGl2YXRlIFRlc3RTdWl0ZVxuICAgIFRlc3RDYXNlU3RhcnRBbmRTdG9wIC0-PiBSb2JvdDogc3RhcnRcbiAgICBhY3RpdmF0ZSBSb2JvdFxuICAgIFRlc3RDYXNlU3RhcnRBbmRTdG9wIC0-PiBSb2JvdDogc3RvcFxuICAgIGRlYWN0aXZhdGUgUm9ib3RcbiAgICBkZWFjdGl2YXRlIFRlc3RDYXNlU3RhcnRBbmRTdG9wXG4gICAgVGVzdFN1aXRlIC0-PiBUZXN0UmVzb3VyY2U6IHNldFVwXG4gICAgVGVzdFJlc291cmNlIC0-PiBSb2JvdDogc3RhcnRcbiAgICBhY3RpdmF0ZSBSb2JvdFxuICAgIFRlc3RTdWl0ZSAtPj4gK1Rlc3RDYXNlQWN0aW9uczogcnVuXG4gICAgVGVzdENhc2VBY3Rpb25zIC0-PiBSb2JvdDogYWN0aW9uMVxuICAgIFRlc3RDYXNlQWN0aW9ucyAtPj4gUm9ib3Q6IGFjdGlvbjJcbiAgICBkZWFjdGl2YXRlIFRlc3RDYXNlQWN0aW9uc1xuICAgIFRlc3RTdWl0ZSAtPj4gVGVzdFJlc291cmNlOiB0ZWFyRG93blxuICAgIFRlc3RSZXNvdXJjZSAtPj4gUm9ib3Q6IHN0b3BcbiAgICBkZWFjdGl2YXRlIFJvYm90XG4gICAgZGVhY3RpdmF0ZSBUZXN0U3VpdGUiLCJtZXJtYWlkIjp7fSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

But if you don't implement a composite structure to your testSuites and you aggregate all testCases into one single testSuite this **_actually_** will look like this:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVGVzdFN1aXRlXG4gICAgcGFydGljaXBhbnQgVGVzdENhc2VTdGFydEFuZFN0b3BcbiAgICBwYXJ0aWNpcGFudCBUZXN0Q2FzZUFjdGlvbnNcbiAgICBwYXJ0aWNpcGFudCBUZXN0UmVzb3VyY2VcbiAgICBwYXJ0aWNpcGFudCBSb2JvdFxuICAgIFRlc3RTdWl0ZSAtPj4gVGVzdFJlc291cmNlOiBzZXRVcFxuICAgICAgICBhY3RpdmF0ZSBUZXN0U3VpdGVcbiAgICBUZXN0UmVzb3VyY2UgLT4-IFJvYm90OiBzdGFydFxuICAgIGFjdGl2YXRlIFJvYm90XG4gICAgVGVzdFN1aXRlIC0-PiArVGVzdENhc2VTdGFydEFuZFN0b3A6IHJ1blxuXG4gICAgVGVzdENhc2VTdGFydEFuZFN0b3AgLT4-IFJvYm90OiBzdGFydFxuICAgIFRlc3RDYXNlU3RhcnRBbmRTdG9wIC0-PiBSb2JvdDogc3RvcFxuICAgIGRlYWN0aXZhdGUgUm9ib3RcbiAgICBkZWFjdGl2YXRlIFRlc3RDYXNlU3RhcnRBbmRTdG9wXG4gXG5cbiAgICBUZXN0U3VpdGUgLT4-ICtUZXN0Q2FzZUFjdGlvbnM6IHJ1blxuICAgIFRlc3RDYXNlQWN0aW9ucyAtPj4gUm9ib3Q6IGFjdGlvbjEgLS0tIHdpbGwgZmFpbFxuICAgIFRlc3RDYXNlQWN0aW9ucyAtPj4gUm9ib3Q6IGFjdGlvbjIgLS0tIHdpbGwgZmFpbFxuICAgIGRlYWN0aXZhdGUgVGVzdENhc2VBY3Rpb25zXG4gICAgVGVzdFN1aXRlIC0-PiBUZXN0UmVzb3VyY2U6IHRlYXJEb3duXG4gICAgVGVzdFJlc291cmNlIC0-PiBSb2JvdDogc3RvcCAtLS0gbWlnaHQgZmFpbFxuICAgIGRlYWN0aXZhdGUgVGVzdFN1aXRlIiwibWVybWFpZCI6e30sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)


Because all necessary test resources are gathered by the suite to be setup and teardown before and after running the tests, this usecase will systematically fail when executing both test cases in the same testSuite. 

Obviously the robot example is only one example, it could be any type of complex system (singleton like) that takes several minutes to start up (web server + microservices, a  test database) -- something you don't want to start before every tests.

mattonem:
> I'm afraid I was expecting that. However, the class comment of Squeak's `TestSuite` mentions:
> _... The common protocol is #run: aTestResult and the dependencies protocol_
> 
> and `#run:` calls `setUp` and `tearDown` correctly.
> 
> >

I'm afraid it does not call testSuite level `setUp` and `teadDown` properly.


`TestSuite>>setUp` basically does not exist in Squeak... 
`setUp` and `teadDown` are only supported at TestCase level...

Not saying it bad or a bug, but just pointing out a major difference in the Sunit framework in both world which is driving me crazy... 

for reference, the piece of code in both language that would call `TestSuite>>setUp`.

- in **_squeak_**:
```smalltalk
run: aResult
      self tests do: [:each | 
		self changed: each.
		each run: aResult].
```
- and **_Pharo_**:
```smalltalk
runUnmanaged: aResult
	self setUp.
	[ self tests
		do: [ :each | 
			each run: aResult.
			self announceTest: each.
			self changed: each ] ]
		ensure: [ self tearDown ]
```

fniephaus:
Ok, in that case I suggest you focus on implementing this in the Pharo-specific packages, at least for now. Also note that you can build and test your projects with a specific branch or fork of smalltalkCI (see [here](https://github.com/marketplace/actions/setup-smalltalkci#use-a-different-branch-or-fork) for example), which may be handy to verify it works correctly for your use cases.

marceltaeumel:
Since `TestSuite` already checks the availability before and resets test resources after, it makes sense to generalize this behavior into a `#setUp`/`#tearDown` combo like the ones in regular tests.

PR #512, opened 2021-04-12
We should use TestSuite composite structure in Pharo

mattonem:
following up PR #509


mattonem:
tested on project [browserstack-local-smalltalk](https://github.com/mattonem/browserstack-local-smalltalk/runs/2325657703?check_suite_focus=true) which requires parallel testing & singleton server  that requires to be initialised at the the beginning of the test suite.

I will also add a testCase in the smalltalkCI test suite.

mattonem:
I must admit, Im a bit confused with Moose failing here... 
- Using `./bin/smalltalkCI --headfull` to start a Moose image, all the code is not loaded into the image... only the latest version of the test package is loaded and the master version of SmalltalkCI-Pharo-Core...
- Whereas Using `./bin/smalltalkCI --headfull` to start a Pharo image, all the latest code is loaded...

fniephaus:
When you run with `--debug`, a debugger should be opened that let's you understand what's happening during bootstrapping and testing. @badetitou confirmed that Moose is built with smalltalkCI and because of that, it contains the version of smalltalkCI used at build-time. I have the feeling that https://github.com/hpi-swa/smalltalkCI/commit/60313510cd56faeef72babb80d2413977ee5396d is not enough to actually override the loaded smalltalkCI version. That's where I'd start looking for the issue...

badetitou:
I confirm again, the Moose image are built with SmalltalkCI.
The GitHub Actions are here for Moose 8 -> https://github.com/moosetechnology/Moose/blob/v8/.github/workflows/continuous-v8.yml (and it is a continuous release on the `v8.x.x` tag)
for Moose 9 -> https://github.com/moosetechnology/Moose/blob/development/.github/workflows/continuous.yml (and it is a continuous release on the `continuous` tag)

mattonem:
Is it possible that because I've made a couple of those changes using github online editor, I ve not updated the methodsProperties.json files in the core package. Hence, Metacello might not detect that a new version of the package is to be loaded....

fniephaus:
>Is it possible that because I've made a couple of those changes using github online editor, I ve not updated the methodsProperties.json files in the core package. Hence, Metacello might not detect that a new version of the package is to be loaded....

At least in Squeak, Metacello doesn't know about the version number of code checked in with FileTree, it's always `cypress.1` or something like that. But that's something you could easily check with `--debug --headless`.

mattonem:
That was it... when upgrading packages, metacello/monticello will compare the versions from the metadata file...  
Conclusion, don't use Github live editor...

mattonem:
and now gemstone won't load...

mattonem:
Actually, I really have no expertise with gemstone... and I wouldnt know how to debug that one. 
Looks like a network connexion failure... Have no idea what it means. Any help with that one would be welcome.

fniephaus:
GemStone sometimes fails in obscure ways, and often because some incompatible code has been added to global packages. In this case, I assume that something in your new `SCITestRunner>>totalTests` does not work in GemStone (e.g. `valuesDo:`). @dalehenrich would probably know more. Is there any way you could move that implementation into the Pharo package as well? Maybe that's all you have to do to make the GemStone build pass.

mattonem:
> I assume that something in your new `SCITestRunner>>totalTests` does not work in GemStone

I'm not sure about that. As `SCITestRunner>>erroredTest` uses exactly the same implementation

fniephaus:
Oh, looks you are right. I've restarted the CI builds a third time. Maybe the build is just a bit flaky....we'll see.

fniephaus:
@dalehenrich any idea what's going on here? See:
https://github.com/hpi-swa/smalltalkCI/pull/512/checks?check_run_id=2363548197#step:4:1355

fniephaus:
The GemStone failures are unrelated (see https://github.com/hpi-swa/smalltalkCI/issues/514). I'll try to get this merged asap...

fniephaus:
Sorry for the delay, cleaning up the metadata took longer than expected. Thanks again for your contributions!

PR #517, opened 2021-04-26
download_file(): Raise timeouts

### LINKED ISSUE #516: curl: (28) Operation timed out, failed to download Squeak Trunk image
While debugging smalltalkCI locally, I received the following error multiple times:

```
curl: (28) Operation timed out after 29979 milliseconds with 26840721 out of 29105923 bytes received
curl failed to download https://github.com/hpi-swa/smalltalkCI/releases/download/v2.9.6/Squeak64-trunk-20447.tar.gz to '/path/to/smalltalkCI/_builds/2021_04_26_18_03_43/Squeak64-trunk-20447.tar.gz'.
```

Since the download was already completed by 91% in this case, would you agree to increase the timeout?

LinqLover:
Fixes #516. Note that in case of a connection error such as 404, this timeout would not be relevant but we would receive an immediate error instead.

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -324,11 +324,11 @@ download_file() {
 
   if program_exists "curl"; then
     curl --fail --silent --show-error --location \
-      --retry 3 --retry-delay 5 --max-time 30 \
+      --retry 3 --retry-delay 5 --max-time 300 \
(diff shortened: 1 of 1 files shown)


fniephaus (approved):
LGTM, thanks!

PR #520, opened 2021-05-11
CLI: Fix --clean option if configuration is absent

### LINKED ISSUE #519: Unable to invoke smalltalkCI with option `--clean` if `.smalltalk.ston` is absent
- Steps to reproduce: `cd smalltalkCI/bin && ./smalltalkci --clean`
- Expected behavior: Clean up works as usual
- Actual behavior: `No STON file named '.smalltalk.ston'`

The problem appears to be that in the `run.sh`, `ensure_ston_config_exists` is called before `check_clean_up`:

https://github.com/hpi-swa/smalltalkCI/blob/cac96a5fd62f38b04dfacf774baec2fa946f0f97/run.sh#L525-L529

Intuitively, I would have swapped the two last statements, but to support continuation after the `--clean` option, `check_clean_up` depends on `config_ston` which is initialized in `ensure_ston_config_exists` only:

https://github.com/hpi-swa/smalltalkCI/blob/cac96a5fd62f38b04dfacf774baec2fa946f0f97/run.sh#L442-L446

Would it be okay to check for `config_first_arg_or_empty` instead to eliminate this dependency?

LinqLover:
`run.sh`: In `main()`, invert the invocation order of `check_clean_up` and 
`ensure_ston_config_exists`, to prevent error messages regarding the absence of configuration when the command was invoked to clean up the cache only. Resolve the existing dependency order which was manifested by the 
`config_smalltalk` variable (read in `check_clean_up` and written in `ensure_ston_config_exists`) by moving the "Continue with build progress?" prompt into the latter function, `ensure_ston_config_exists`. We only show this prompt if the local variable 
`first_action` has been set, which at the moment only happens in `check_clean_up`.

Fixes #519.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -9,6 +9,10 @@ readonly BINTRAY_API="https://api.bintray.com/content"
 readonly DEFAULT_STON_CONFIG="smalltalk.ston"
 readonly GITHUB_REPO_URL="https://github.com/hpi-swa/smalltalkCI"
 
+# Indicates whether any secondary action was chosen that makes the execution
+# of the actual build process optional. See ensure_ston_config_exists.
+first_action=""
+
 ################################################################################
 # Locate $SMALLTALK_CI_HOME and load helpers.
 ################################################################################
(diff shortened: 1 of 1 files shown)


fniephaus (approved):
LGTM

PR #522, opened 2021-05-16
Fuel pharo9 fixes

theseion:
Fixes one deprecation warning and the use of FuelOutStackDebugAction.


theseion:
Note that the change is not yet in Pharo 9 but shouldn't be far off. See https://github.com/pharo-project/pharo/issues/9282

fniephaus:
Thanks, @theseion! Looks like you deleted [`repository/SmalltalkCI-Pharo-Core.package/monticello.meta/version` and GemStone doesn't like that.](https://github.com/hpi-swa/smalltalkCI/pull/522/checks?check_run_id=2594965106#step:4:1177)

theseion:
Argh! Thank. I'll fix that.

fniephaus (approved):
LGTM, thanks!

theseion:
Thanks for the fixes. I mess up with Pharo vs. Filetree every time 🙄 .

fniephaus:
No worries! I wish we could nuke all metadata files... but unfortunately, this would break compatibility with older versions of Squeak and Pharo :(

PR #526, opened 2021-05-21
Before using Metacello>>onConflictUseIncoming, check if its there.

theseion:
This method doesn't exist in the Metacello versions loaded into Pharo 3 and 4. Instead, we can use #onConflict: directly

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -304,15 +304,20 @@ pharo::load_project() {
   pharo::run_script "
     | smalltalkCI |
     $(conditional_debug_halt)
-    [ Metacello new
-        baseline: 'SmalltalkCI';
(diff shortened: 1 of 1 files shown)


fniephaus:
Mh, should we maybe add a Pharo 3 build to our [build matrix](https://github.com/hpi-swa/smalltalkCI/blob/ba92e0ca58e5c189289ed19ae49a8b6bd6544c83/.github/workflows/main.yml#L22-L36)?

theseion:
Depends... I seem to be the only one testing those :). My two cents: do it if it's a simple change and doesn't add any significant build costs. Otherwise I wouldn't bother.

fniephaus:
I've added a couple of more images to the build matrix: https://github.com/hpi-swa/smalltalkCI/commit/9bbacfa012d8ec7e00996c880147af51ff42d573

Should be ok for now :)

PR #528, opened 2021-05-24
No longer load STON for Squeak >= 5.4

### LINKED ISSUE #527: SmalltalkCI overrides STON-Core in Squeak trunk
Squeak trunk (5.4) can't currently load any code. The reason for this appears to be a name clash of `STON-Core`. Squeak now comes with its own version of STON (https://github.com/squeak-smalltalk/squeak-ston) which includes a class names `STONJSON` which is need to load code from Tonel repositories. SmalltalkCI inadvertently replaces this package, thus removing `STONJSON`.

I suspect the SmalltalkCI baseline needs to be updated for 5.4 to exclude the `STON` package.

theseion:
Split package and group methods for Squeak in two (similar to Pharo).

Fixes #527.


tom95:
Thank you for the PR @theseion! Two things to note still:
* there will most likely not be a squeak5.4, only squeak6
* I have a hunch that squeak<=5.3 builds would also break if their cached images were ever updated to include the newer Metacello. A more longterm fix for this odd dependency incompatibility issue might thus be actually a check whether STON is already in the image instead of dispatching based on version.

But I'm not a 100% sure on the second point just yet. Is this issue in any way time-critical for you? Otherwise I would check laeter today or tomorrow (unless you beat me to it of course :)) and report back, if successful maybe with a new PR based on yours.

theseion:
> SmalltalkCI overrides STON-Core in Squeak trunk
True. But 6 has been coming for about 2 years now? And 5.4 matches the Metacello attribute if I'm not mistaken.

> I have a hunch that squeak<=5.3 builds would also break
Right. I wasn't aware that's an option. 
We could just put that check into the baseline, like you suggested. I think that would be fine. There is one alternative though and that is to depend on a specific Metacello version in the SmalltalkCI baseline. I don't think we should do that but I'm pointing it out for completeness.

This isn't time critical but would be nice to get fixed in the next couple of days (or weeks). Feel free to adapt the PR.

theseion:
Superseded by #530.

PR #530, opened 2021-05-25
Squeak: use up-to-date STON, remove bundled STON

tom95:
Addresses https://github.com/hpi-swa/smalltalkCI/issues/527
Supersedes https://github.com/hpi-swa/smalltalkCI/pull/528

/cc @theseion 

I split the change into two commits for a little more overview. The first adapts the baseline for squeak to load the up-to-date STON package. The second deletes the bundled STON package.

Locally, I was only able to test the 64bit images (so starting with 5.1), so I'm curious to see what CI will say for squeak4.x :)

As far as I've seen, neither Gemstone nor Pharo depend on the bundled STON, at least it is not referenced in any place that I found.


fniephaus:
> [error] The action has timed out.

Not sure what is causing this, and it seems I can no longer reproduce this locally as I'm on a Mac 😬

fniephaus (approved):
LGTM, thanks!

fniephaus:
@theseion does this work for you?

theseion (approved):
Looks good. I've tested with Squeak32-4.6 too. Nice idea to load STON from github +1.

PR #531, opened 2021-05-26
Updated appveyor.yaml example in README.md to use 64-bit Cygwin

theseion:
My Fuel builds on AppVeyor have been failing because I was using the wrong architecture. I had not realized that Cygwin must be bootstrapped to 64-bits.

I have updated the appveyor.yaml example in the README.md to use 64-bit Cygwin, as the images in that example are 64-bit anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,11 +164,12 @@ smalltalk:
 ### `appveyor.yml` Template
 
 ```yml
+# This is a 64-bit environment, so use 64-bit Cygwin
 environment:
-  CYG_ROOT: C:\cygwin
-  CYG_BASH: C:\cygwin\bin\bash
-  CYG_CACHE: C:\cygwin\var\cache\setup
-  CYG_EXE: C:\cygwin\setup-x86.exe
+  CYG_ROOT: C:\cygwin64
+  CYG_BASH: C:\cygwin64\bin\bash
+  CYG_CACHE: C:\cygwin64\var\cache\setup
+  CYG_EXE: C:\cygwin64\setup-x86_64.exe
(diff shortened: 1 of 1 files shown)


fniephaus (approved):
LGTM, thanks!

fniephaus:
Do we have to update smalltalkCI's appveyor.yml? Why does mixing 32 and 64 bit architectures not lead to problems there?

theseion:
I'm not sure. Maybe specifying
```yaml
platform
  - x64
```
has some effect, since the host architecture is 64-bits? In any case, I'm not sure there where *no* issues, I did see frequent segfaults. I just always attributed those to the VM...

I didn't change the appveyor.yml precisely because of the platform mix. I'm not sure how to handle that, since we'd probably need to switch environments based on the build, e.g. `Squeak32-5.1` -> use `CYG64`. I only have trunk and alpha running on AppVeyor, so I guess it depends on how many people actually need it to work with mixed architectures...

fniephaus:
I always thought that 32-bit "just work" on 64-bit Windows. Maybe @marceltaeumel knows more? Why would the arch of Cygwin matter in the first place?

theseion:
Well, in the case of Pharo 9+ that may be true but won't matter because the VM file name contains the architecture and there are no longer any 32-bit VMs for Pharo 9+. The reported architecture for 32-bits is `i686`. The situation is a bit messy. I did in fact try [to contribute a fix to zeroconf](https://github.com/pharo-project/pharo-zeroconf/pull/33) to fix the problem but it's not that simple. I could revise the change I've already made to fall back to x86_64... that would probably work reasonably well...

PR #535, opened 2021-05-31
remove not used attributes

estebanlm:
sorry, in my last somehow I lost this change (the mess with the properties) :(
I am removing two unused attributes : 
```
SmalltalkCI (coverageWrappers)
SCITestRunnerResult (error)
```

(sorry again for the involuntary property mess)


estebanlm:
mmm.... error looks like some timeout?

estebanlm:
@fniephaus, any chance with this?

fniephaus:
Done! How can we make sure that this doesn't happen again in smalltalkCI? Could you maybe contribute some appropriate tests?

estebanlm:
I can contribute the release tests of Pharo ;)
the class ReleaseTests verify all that... I will do a subset and make a PR ;)

fniephaus:
Thanks! How long do these `ReleaseTests` run? Maybe, they can just be invoked as part of smalltalkCI's test config.

estebanlm:
the part that stCI can use it, will not take much time to run... I guess. I will check :)

fniephaus:
Would be nice if we could avoid duplicating this part of code, so we don't have to keep anything in sync :)

estebanlm:
well, `ReleaseTests` is the quality package that runs as the final part of Pharo and it gets improved all the time, I really do not see possible to move tests from Pharo to stCI... I guess we will need to live with the duplication here :/

fniephaus:
That's unfortunate. We do run a couple of Gemstone-specific tests with smalltalkCI: https://github.com/hpi-swa/smalltalkCI/blob/f10f4e055186d960b27aa0ac4b4ae28e850e3fcb/.smalltalk.ston#L19

PR #538, opened 2021-07-21
update to Pharo 10

estebanlm:
We opened Pharo 10 development, which means we need to update smalltalkCI for it :)


fniephaus:
Has the versioning scheme changed or why is there no minor version number anymore (Pharo 10 vs 10.0)?

estebanlm:
yep, no minor version because it has no sense (we do not use it anymore)... :)
now, I can use it anyway if needed (since *eventually* could be used), just we want to simplify :P

estebanlm:
where are this test_get_image_url and test_get_vm_url located? 
I need to update them...
also I am not sure they are good tests, since stable change a lot (once a year, average) ;)

fniephaus:
>also I am not sure they are good tests, since stable change a lot (once a year, average) ;)

Sure, maybe we should remove the tests based on stable/alpha tags. the rest should never change, right?

estebanlm:
indeed :)

fniephaus:
> Unsupported Pharo version 'Pharo64-stable'.

That's indeed a valid bug, you removed `Pharo64-stable` from the vm map.

estebanlm:
mmm... need to go now, I'll be back to this the afternoon

estebanlm:
btw, I was mistaken... they are actually very good tests, I would have miss some targets without them :)

fniephaus:
> btw, I was mistaken... they are actually very good tests, I would have miss some targets without them :)

They are quite annoying to work with if you run them on CI only... but when you run them locally, they are super fast and avoid bugs quite well. :)

fniephaus (approved):
LGTM, only a `Pharo64-10` build job is missing in the GitHub workflow...will add one after merging. Thanks, @estebanlm!

PR #543, opened 2021-12-27
Fix invocation of image/VM with spaces in path

LinqLover:
Without this patch, it is not possible to pass a VM path containing spaces to smalltalkCI, like here in mingw:

```
$ smalltalkci -s Squeak64-Trunk --vm "/c/Program Files (x86)/Squeak/SqueakConsole.exe"
```

Analogously, you previously could not pass an image path containing spaces like

```
$ smalltalkci -s Squeak64-Trunk --image "/path/to/My Squeak.image"
```

For more information, see: https://github.com/koalaman/shellcheck/wiki/SC2124


LinqLover:
By the way, there are loads of other warnings caused by shellcheck in all the script files in this repository. It would probably not harm to address them at some later point in time as well. 😅

fniephaus (approved):
LGTM. thanks!

fniephaus:
> By the way, there are loads of other warnings caused by shellcheck in all the script files in this repository. It would probably not harm to address them at some later point in time as well.

Yeah, I agree. Would be nice to add a shellcheck build job at some point.

fniephaus:
Integrated as part of #575.

PR #548, opened 2022-02-17
Add Moose 10

ClotildeToullec:
- Added Moose 10 to available Moose versions
- Updated Moose 9 and Moose 8 urls

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,9 +37,9 @@ jobs:
           - Pharo32-stable
           - Pharo32-alpha
           - Pharo32-3.0
+          - Moose64-10
           - Moose64-9.0
           - Moose64-8.0
-          - Moose64-7.0
         exclude: # exclude 32bit builds on macOS
           - os: macos-latest
             smalltalk: Squeak32-trunk
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -91,7 +91,7 @@ moose::get_image_url() {
 
   case "${smalltalk_name}" in
     "Moose64-trunk"|"Moose-trunk")
-      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose9-development.zip"
+      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose10-development-Pharo64-10.zip"
       ;;
     "Moose32-trunk")
       moose_name="moose-7.0"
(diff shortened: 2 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -37,9 +37,9 @@ jobs:
           - Pharo32-stable
           - Pharo32-alpha
           - Pharo32-3.0
+          - Moose64-10
```
fniephaus, line 40: For consistency, should this rather be `Moose64-10.0` or has the version scheme be simplified for Moose?

.github/workflows/main.yml
```diff
@@ -37,9 +37,9 @@ jobs:
...
           - Moose64-9.0
           - Moose64-8.0
-          - Moose64-7.0
```
fniephaus, line 42: Let's drop `Moose64-8.0` instead of `Moose64-7.0` so that we still test the oldest supported version.

pharo/run.sh
```diff
@@ -106,10 +106,13 @@ moose::get_image_url() {
...
     "Moose64-8"*)
-      echo "https://github.com/moosetechnology/Moose/releases/download/v8.x.x/Moose8-stable.zip"
+      echo "https://github.com/moosetechnology/Moose/releases/download/v8.x.x/Moose8-old-stable-Pharo64-8.0.zip"
       ;;
     "Moose64-9"*)
-      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose9-development.zip"
+      echo "https://github.com/moosetechnology/Moose/releases/download/v9.x.x/Moose9-stable-Pharo64-9.0.zip"
+      ;;
+    "Moose64-10"*)
+      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose10-development-Pharo64-10.zip"
```
fniephaus, line 115: Looking at the URL, the version scheme might indeed have changed. Could you please clarify?

fniephaus (approved):
Thanks for the contribution! I left two minor remarks but otherwise, this looks good.

.github/workflows/main.yml
ClotildeToullec: We are following Pharo scheme (Pharo64-9.0 then Pharo64-10) but this is not mandatory at all. We can change it if necessary for SmalltalkCI. However, I would like not to change the url following this scheme.

.github/workflows/main.yml
ClotildeToullec: Moose 7 is not supported anymore. Moose 8 is.

fniephaus: Ok, good!

.github/workflows/main.yml
fniephaus: Oh right, I forgot about `Pharo64-10`. Thanks!

PR #549, opened 2022-02-22
add pharo10

estebanlm:
baseline is not working for pharo10


estebanlm:
> Thanks! I wonder how this has ever worked for Pharo-10?

because when developing P10, we keep it "metacello-compatible" with P9 to easy things. But now we are getting ready to release :)

fniephaus:
The GitHub Action appears to have some problems at the moment. I've merged this anyway. Thanks

Issue #362, opened 2018-03-31
Coveralls testing is skipping class side & extension methods


peteruhnak:
Apparently class-side and extension methods are both being skipped during coverage collection.

Here's a sample [coveralls_results.json](https://gist.github.com/peteruhnak/675cd4515d457c91074fa03455af3b89)

here's the build https://travis-ci.org/peteruhnak/pharo-travis-pipeline/jobs/360667859

Testing itself works, fine, but not collection.

E.g. [PTItem class>>testedClassSideMethod](https://github.com/peteruhnak/pharo-travis-pipeline/blob/master/repository/PharoTravisPipeline.package/PTItem.class/class/testedClassSideMethod.st) is correctly tested (Line 735 [#testClassSideMethod](https://travis-ci.org/peteruhnak/pharo-travis-pipeline/jobs/360667859#L735))

but there's no mention of `testedClassSideMethod` in the JSON (instead there's just `null`)

Dtto for extension methods https://github.com/peteruhnak/pharo-travis-pipeline/tree/master/repository/PharoTravisPipelineExtensions.package

LinqLover:
I could reproduce this issue with multiple Squeak repositories for each of which the Squeak Test Runner displays the coverage results correctly:

![image](https://user-images.githubusercontent.com/38782922/147787524-9f6797b5-1155-46b0-bf53-1d2fa17b51df.png)
![image](https://user-images.githubusercontent.com/38782922/147787540-934e52ed-f107-40ff-bb12-35e5164eb23e.png)

The following kinds of methods are not written correctly into the JSON file:

- class-side methods
- extension methods
- trait methods (neither registered in the trait behavior nor in its users)
- special selectors such as `#~~`

The defect method is `SCICodeCoverage class >> #filePathFor:in:`, which uses a probably too simple heuristic for creating the FileTree paths for all methods:

https://github.com/hpi-swa/smalltalkCI/blob/b30166bda3aabf8c74266f6fbaf85b7592891cdd/repository/SmalltalkCI-Coverage-Core.package/SCICodeCoverage.class/class/filePathFor.in..st#L1-L16

Fixing class-side methods is easy, just exchange `class name` with `class theNonMetaClassName`.  
For extension methods, `#packageNameForClass:` needs to be replaced with something that takes the extension category of the method into consideration.  
For traits, things get more complicated because afaik all compiled methods from a trait are copied into the method dictionary of all of its users.  
[...]

PR #565, opened 2022-09-09
Fix file paths for coverage reports

LinqLover:
Implement and test proper handling of class-side methods, extension methods, trait methods, and special message selectors.

Resolves https://github.com/hpi-swa/smalltalkCI/issues/362.

_Draft: I still need to test this manually and add pharo compatibility._


LinqLover:
(oops, I did not assume that renaming a branch via GitHub would close all associated PRs ... Opened #571 instead.)

PR #566, opened 2022-09-09
Add `workflow_dispatch` trigger to CI

LinqLover:
Make it possible to trigger the CI manually from the Actions page.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,7 @@ on:
       - 'docs/**'
       - '.appveyor.yml'
       - '.travis.yml'
+  workflow_dispatch:
 
 jobs:
   build:


fniephaus:
Should this be mentioned in the setup-smalltalkCI templates as well?

fniephaus (approved):

LinqLover:
> Should this be mentioned in the setup-smalltalkCI templates as well?

Good question, I don't know. Since `workflow_dispatch` is a general GitHub Actions concept and not related to this specific action, I would tend against it. What do you think?

But FWIW, I added it to the SWT-Demo config (https://github.com/hpi-swa-teaching/SWT-Demo/commit/a167a3c766e21af8ca947158469394f920f98d27).

PR #578, opened 2022-10-14
Fix CLI option `--self-test` to be used in combination with other arguments

LinqLover:
Previously, any arguments after `--self-test` were swallowed, so for instance, the following did not have the intended effect:

    bin/smalltalkci --self-test --headfull

This patch resolves this limitation.

I wonder whether this was a slip or intended ... But I can't imagine any scenario in which ignoring all further arguments would be intended. At last, there is also `#` in any bash shell to do this.


fniephaus (approved):
Thanks for fixing, I think it's a bug not a feature.

Issue #552, opened 2022-04-26
[Pharo] The tested repository does not register itself with Iceberg


hogoww:
The image created for Pharo testing does not register the repository on Iceberg.
Because of that, Iceberg features cannot be used.

Particularly, I generally rely on Iceberg to tell me where the repository is in the file system.
Which allows me to use external resources.
(Maybe there's a portable equivalent, but I only use iceberg :/).

I don't have time to investigate it myself, I fixed it in another way for my project.

Pierre

fniephaus:
Not sure what smalltalkCI would need to do. Could an Iceberg user comment on this? Maybe @gcotelli or @estebanlm?

fniephaus:
While we wait for someone knowledgeable, I'd like to point out that smalltalkCI supports [custom scripts](https://github.com/hpi-swa/smalltalkCI#custom-scripts), which you could use as a temporary workaround.

fniephaus:
Fixed with #580.

PR #579, opened 2022-11-10
[improvement] add Iceberg registering for the loaded project

hogoww:
This fixes the issue #552 I opened a while back.
Let me know if I have to do more :)

Pierre


fniephaus:
Thanks, @hogoww. Could you please remove all unrelated metadata changes so we can better see what has actually changed?

So far, I've only seen [SmalltalkCI.class/instance/load.st](https://github.com/hpi-swa/smalltalkCI/pull/579/files#diff-c402f23cc70ffbc559d5218bc6a509825de472ae38e8cddfb547efef713c0bf7), which is unfortunately in the generic `SmalltalkCI` class. I'd assume this only works in Pharo so it should probably be moved to one of the classes in the [SmalltalkCI-Pharo-Core.package](https://github.com/hpi-swa/smalltalkCI/tree/master/repository/SmalltalkCI-Pharo-Core.package), for example `SmalltalkCIPharo11`? Maybe @estebanlm has an opinion on that?

hogoww:
Oh my.
I was clearly not well woken up when I did this.
Sorry, I'll redo it properly a bit later :D
Thanks @fniephaus !

Pierre

PR #580, opened 2022-11-10
[improvement] now register the tested repository in iceberg for P9+

hogoww:
Clean redo of #579 fixing #552 !
I've enabled the registering of the repository for Pharo 9+ (therefore 10 and 11 at this time).

Pierre


theseion (changes requested):
I like the idea!

While the change looks good, I would like to have it wrapped in an error handler. SmalltalkCI might run on anything, we can't be certain that this will be a valid Pharo repository. If it's not, the tests should still be run.

I also don't like the override too much. I think it would make sense to have an `afterLoad` method that can be overridden and would do error handling automatically.

hogoww:
> I like the idea!
 :) 

> While the change looks good, I would like to have it wrapped in an error handler. SmalltalkCI might run on anything, we can't be certain that this will be a valid Pharo repository. If it's not, the tests should still be run.

I'd have to check, but I think as long as it's a valid Monticello repository, it's valid for Iceberg.
Would that be a fair assumption to have ?
Still, I think the error handler is a good idea regardless.

> I also don't like the override too much. I think it would make sense to have an `afterLoad` method that can be overridden and would do error handling automatically.

I like that too !
Will do :)

Another thing I was planning to add in a next PR would be to add the SmalltalkCI repository to Iceberg aswell, in the pharo execution script.
That'd be easier for other purposes (such as creating a PR !)

Pierre

theseion:
> Another thing I was planning to add in a next PR would be to add the SmalltalkCI repository to Iceberg aswell, in the pharo execution script.

Might be a bit harder though, since you don't know where that repo is in relation to the one you're testing. Or do you mean to add it as a new clone?

hogoww:
Actually, since you start from a Pharo script, it's fairly easy.
That is, if my quick morning investigation was done properly ;)

theseion:
Great then :)

hogoww:
Icerberg seems to commit metadata changes silently...
Anyway, should be fine now.
@theseion does that fit your requirements? 
I wouldn't be against a better error message IMO.

repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo9.class/instance/postLoad.st
```diff
@@ -0,0 +1,8 @@
+executing
+postLoad
+	[ (IceRepositoryCreator new
+		location: projectDirectory asFileReference;
+		createRepository) register. ] 
+		on: Error 
+		do: [ :e | self class printWarningToStdout: 'Could not register the tested repository to Iceberg for some reason.'.
```
theseion, line 7:
```suggestion
		do: [ :e | self class printWarningToStdout: 'Could not register the tested repository with Iceberg.'.
```

theseion (changes requested):

theseion (approved):

theseion:
@fniephaus You can merge this PR, unless you want to wait for @gcotelli's review.

fniephaus (approved):
LGTM, thanks!

PR #582, opened 2022-11-20
Iceberg registering in github actions

hogoww:
Hello,

This PR "fixes" #581 by redirecting users to the setup-SmalltalkCI link on which I will do another PR to add the requirements to make Iceberg work in github actions.
Basically, we need to fetch all of the commits because Iceberg always creates a full history as far as I can tell.
This is achieve with the option fetch-depth: 0 for actions/checkout@v2
```yaml
- uses: actions/checkout@v2
        with:
          fetch-depth: 0
```
(Will do that PR later today)

I also put Iceberg in a preload rather than postload, to be able to use it during the loading.
(In my case, this was necessary for resource usage).
If you think there should be another explanation on how to Iceberg in the CI, let me know which file I should add it to.

Thanks again to @theseion for the support & @fniephaus for the solution !

PR #584, opened 2023-01-16
Redo of #582 + fix

hogoww:
Add a compatibility for people using Pharo but not metacello, following @fniephaus comments!


hogoww:
Anything I need to do ?
Couldn't fix it faster than right when I saw it :p

fniephaus:
Thank you, @hogoww! :)

jvalteren:
I just came across a different use case where I need to access Iceberg repository information during baseline loading. Works like a charm! :-)

Thanks for the hard work guys!

Issue #585, opened 2023-01-22
Code coverage in Pharo is unexpectedly wrapping all methods of a class that is extended by a package under test


jbrichau:
When a Pharo package that is counted for code coverage extends a class that is not in the package, all methods inside that class are wrapped with the `SCICoverageWrapper` anyway. If this occurs with a method on `UndefinedObject`, the entire Pharo image becomes unresponsive. 

For example: Grease extends the class `UndefinedObject`. Because of this, SmalltalkCI is wrapping all methods in `UndefinedObject`. The Pharo image becomes unresponsive when executing the following line: https://github.com/hpi-swa/smalltalkCI/blob/44a709d5d133b7e02bbc11a50aba263abb2a813c/repository/SmalltalkCI-Coverage-Core.package/SCICodeCoverage.class/instance/startUp.st#L8z . More specifically, it becomes unresponsive when class methods from `UndefinedObject` are wrapped.

This happens even when specifying the classes to be checked for coverage using `#categories:` rather than `#packages:`.

The origin seems to be that the code in `SmalltalkCIPharo>>classesInPackage:` returns all classes in the package rather than only the classes defined in the package: https://github.com/hpi-swa/smalltalkCI/blob/44a709d5d133b7e02bbc11a50aba263abb2a813c/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo.class/class/classesInPackage..st#L3

When I replace the send of `classes` with the send of `definedClasses` in the method above, the problem goes away.

Mind that the extension methods are still wrapped, as they should, because of the implementation in https://github.com/hpi-swa/smalltalkCI/blob/master/repository/SmalltalkCI-Coverage-Core.package/SCICodeCoverage.class/instance/allMethodReferencesToCover.st

jbrichau:
Related issues:
- https://github.com/hpi-swa/smalltalkCI/issues/495
- https://github.com/hpi-swa/smalltalkCI/issues/476

PR #586, opened 2023-01-22
Fix for #585 (Code coverage in Pharo is unexpectedly wrapping all methods of a class that is extended by a package under test)

jbrichau:
Only consider entire classes defined in a package when retrieving `classesInPackage`.
This fixes https://github.com/hpi-swa/smalltalkCI/issues/585.

Using these modifications, the Grease build works using coverage:
https://github.com/SeasideSt/Grease/actions/runs/3981443794

Before, it used to become unresponsive and timeout:
https://github.com/SeasideSt/Grease/actions/runs/3980634960


jbrichau:
I noticed that committing the entire project via Iceberg modifies and breaks stuff, so I reverted the commit and only selected the necessary method change. 
If the change is accepted, I would recommend doing a squash commit for the PR

fniephaus (approved):
LGTM. Thank you @jbrichau!

PR #587, opened 2023-02-22
Pharo 7 and 8 includes iceberg already and could benefit from the new option

jecisc:
I pushed up the proload from Pharo 9 class to Pharo 78 class


fniephaus (approved):
LGTM, thanks!

PR #593, opened 2023-04-12
update GemStone builds to run on Ubuntu 20.04 and macos-latest

dalehenrich:
The issues:
1.  [GsDevKit_home](https://github.com/GsDevKit/GsDevKit_home) upon which the current version of smalltalkCI for GemStone is built does not function on either Ubuntu 20.04 or the recent MacOS versions, because of it's reliance on 32 bit  Pharo 3.0.
2. GemStone 3.5.3 is the earliest GemStone version that runs on Ubuntu 20.04.

With this pull request, the functionality formerly supplied by [GsDevKit_home](https://github.com/GsDevKit/GsDevKit_home) in the gemstone/run.sh script is now provided by [GsDevKit_stones](https://github.com/GsDevKit/GsDevKit_stones#gsdevkit_stones) the heir apparent to GsDevKit_home and [superDoit](https://github.com/dalehenrich/superDoit#superdoit) a framework for writing shell scripts using GemStone/64 smalltalk that run against an existing stone or against a [standalone topaz solo extent](https://downloads.gemtalksystems.com/docs/GemStone64/3.6.x/GS64-Topaz-3.6/GS64-Topaz-3.6.htm?https://downloads.gemtalksystems.com/docs/GemStone64/3.6.x/GS64-Topaz-3.6/1-Tutorial.htm#pgfId-1131973).

The Ubuntu 20.04 restriction is imposed by GitHub, this new version of smalltalkCI can be run against GemStone versions as old as 3.4.0 on Ubunutu 18.04


helpers.sh
```diff
@@ -550,8 +546,10 @@ report_build_metrics() {
   local duration=$(($(timer_nanoseconds)-$smalltalk_ci_start_time))
   duration=$(echo "${duration}" | awk '{printf "%.3f\n", $1/1000000000}')
 
-  if [[ "${config_tracking}" != "true" ]]; then
-    return 0
+  if [ "${config_tracking+set}" ]; then
```
fniephaus, line 549: Is this still needed?

README.md
```diff
@@ -74,13 +74,13 @@ they can take up a lot of space on your drive.*
 
 | [Squeak][squeak] | [Pharo][pharo]   | [GemStone][gemstone] | [Moose][moose]  | [GToolkit][gtoolkit] | 
 | ---------------- | ---------------- | -------------------- | --------------- | -------------------- |
-| `Squeak64-trunk` | `Pharo64-alpha`  | `GemStone64-3.5.x`   | `Moose64-trunk` | `GToolkit64-release` |
-| `Squeak64-6.0`   | `Pharo64-stable` | `GemStone64-3.4.x`   | `Moose64-10`    |                      |
-| `Squeak64-5.3`   | `Pharo64-11`     | `GemStone64-3.3.x`   | `Moose64-9.0`   |                      |
-| `Squeak64-5.2`   | `Pharo64-10`     | `GemStone64-3.2.x`   | `Moose64-8.0`   |                      |
-| `Squeak64-5.1`   | `Pharo64-9.0`    | `GemStone64-3.1.0.x` | `Moose64-7.0`   |                      |
-| `Squeak32-trunk` | `Pharo64-8.0`    | `Gemstone64-2.4.x`   | `Moose32-trunk` |                      |
```
fniephaus, line 82: This is a breaking change. I'm assuming you're ok with that or should we do something about older GemStone versions?

dalehenrich: I'm not sure what can be done with older versions of GemStone ... they simply fail to run in 20.04 , if when someone hits that the only alternative is to remove the old version from spec ...

helpers.sh
dalehenrich: Not sure ... I [added it](https://github.com/dalehenrich/smalltalkCI/actions/runs/4654504882/jobs/8236306346#step:4:249) because a Moose build was failing, while the Squeak, Pharo, and GemStone builds were not failing ... and that "patch" addressed the problem ... I haven't tracked what you did to "solve" that particular problem in the general case, but if you think it's not needed I can certainly remove it or change it to the "right thing" ...

README.md
fniephaus: smalltalkCI can be used on more than just GitHub Actions, someone may still run it on their own or some other Ubuntu 18.04 box, right?

helpers.sh
fniephaus: I guess those were failing because the subshell that is supposed to upload coverage results failed. Let's revert this and see what the CI says :)

README.md
fniephaus: Oh, maybe we can't re-enable it because the client logic has been removed or does that still work?

helpers.sh
dalehenrich: will do ...

dalehenrich:
I'm not familiar with the set of yaml expressions that are needed to ignore GToolKit failures ... I end up flailing whenever I touch yaml, so if you give me the needed edits I'll make those changes as well and we'll be able to get a full run under our belts ...

helpers.sh
fniephaus: Well, this one failed in `gtoolkit::prepare_gt`, not sure why though.

fniephaus:
>I'm not familiar with the set of yaml expressions that are needed to ignore GToolKit failures ... I end up flailing whenever I touch yaml, so if you give me the needed edits I'll make those changes as well and we'll be able to get a full run under our belts ...

In smalltalkCI, we currently use [this](https://github.com/hpi-swa/smalltalkCI/blob/a370792dd18fbc1f873079fda011f5dccd9dd2ca/.github/workflows/main.yml#L90) for allowed failures. Not sure GitHub has added a better way to specify this.

dalehenrich:
so this would be the expression to use:
```yaml
if: contains(matrix.smalltalk, 'trunk') || contains(matrix.smalltalk, 'alpha') || contains(matrix.smalltalk, 'Squeak32-6.0') || contains(matrix.smalltalk, 'GToolkit64')
```

dalehenrich:
By jove, I think we've got it!

fniephaus (approved):
LGTM, thanks @dalehenrich!

PR #595, opened 2023-04-14
Each time a version of Pharo is released we need to update the baseli…

jecisc:
…ne if we want to be able to be able to load smalltalk CI in the new Pharo version.

Here is a little improvement to not have to do that again.


fniephaus:
Hi @jecisc could you resolve the conflict so we can merge this?

jecisc:
Done

repository/BaselineOfSmalltalkCI.package/BaselineOfSmalltalkCI.class/instance/pharoVersionsFrom..st
```diff
@@ -0,0 +1,4 @@
+baselines
+pharoVersionsFrom: aVersion
+
+	^ (aVersion to: SystemVersion current major) collect: [ :v | ('pharo' , v asString , '.x') asSymbol ]
```
fniephaus, line 4: Turns out this only works in Pharo:
```
a MessageNotUnderstood occurred (error 2010), a Set does not understand  #'current'
```

theseion: Works in Squeak 5.1 as well. Is that error from GS?

fniephaus: I don't know tbh.

fniephaus: Sure, but do we really need to complicate this even more? I forgot about the semantics, but maybe the spec could just target `#'pharo'` instead (or `#'pharoCommon'`?) and then we have specific overrides for Pharo <  5 (e.g. `#'pharo4.x'`)?

jecisc:
Closing since it's more complex than I thought

PR #605, opened 2023-09-21
Propose to change the directory comment

labordep:
I was confuse about this property in my own project. I propose to update the documentation with most used term and to remove the FileTree mention because that's for Tonel too.


fniephaus (approved):
LGTM, thanks!

PR #606, opened 2023-09-23
fix: #listAtCategoryNamed: has been deprecated in Pharo 12

theseion:


fniephaus (approved):
LGTM.

theseion:
Unfortunately, the Pharo stable builds almost always segfault. I didn't manage to get a clean run of all builds.

fniephaus:
> Pharo stable builds almost always segfault

Is this due to a new and fault VM? It's not related to your changes, right? Maybe we need to inform the VM maintainers and mark the test as allowed to fail?

theseion:
I'm pretty sure it's not my fault. Since the Pharo-alpha builds are good, I suspect that the VM used for stable builds really has an issue that has been fixed upstream.

fniephaus:
smalltalkCI uses zeroconf scripts to pull the VM, so this is out of our control. Any suggestion how we could proceed?

theseion:
It's weird. Looking at the builds, the one's for Pharo 11 (explicitly) are fine, the one's for "stable", which is Pharo 11, are not. But since we have both builds anyway, I suggest removing the "stable" builds.

Also, I'm not sure how much sense it makes to test newer versions against the 32-bit VM. We could say that we only support 64-bit explicitly as of Pharo 12 (usually, 32-bit will be fine). That will reduce the number of additional builds in the future

theseion:
Actually, we could conceivably remove all the 32-bit builds. The only ones registered are for Pharo 3, stable and alpha.

theseion:
I've committed a proposal to this PR.

.github/workflows/main.yml
```diff
@@ -26,10 +26,6 @@ jobs:
           - Squeak64-5.3
           - Squeak64-5.2
           - Squeak64-5.1
-          - Squeak32-trunk
-          - Squeak32-6.0
-          - Squeak32-4.5
-          - Pharo64-stable
```
fniephaus, line 32: Why remove `Pharo64-stable` here?

fniephaus:
Looks like you broke the GemStone jobs. Not sure what's going on and not got enough time to help. Sorry!

.github/workflows/main.yml
theseion: 1. Because that's the failing build
2. Because Pharo64-11 is a separate build, already in the list

I'm not saying this is a good suggestion :)

fniephaus:
Do we maybe need to upgrade the stable VMs? https://github.com/hpi-swa/smalltalkCI/blob/d3a45fed245194363720accc47f0655ee9805f4a/pharo/run.sh#L182

theseion:
It's the `SentButNotImplementedTest` from `Rowan` that fails. That's probably because of the Pharo only message that I'm using here: https://github.com/hpi-swa/smalltalkCI/blob/12a3bd332b70265efff5d1ac3cb3ee9144db8a87/repository/SmalltalkCI-Tests.package/SCIMonticelloLoadSpecTest.class/instance/testAddLoadedClassesFrom.st#L12.
Not sure what to do there, except for using dynamic sends to hide the selector...

fniephaus:
I can't remember how this works. Maybe @dalehenrich can help?

theseion:
> Do we maybe need to upgrade the stable VMs?

Oh! Yes, absolutely! I hand't realised that the VMs are hard coded. I'll do that.

theseion:
> I can't remember how this works. Maybe @dalehenrich can help?

It's not a GemStone issue. The issue is that the `SmalltalkCI-Tests` package is loaded for all platforms, but now Pharo is starting to diverge. So maybe we need to start moving those tests to platform specific packages... I'll try to come up with something.

theseion:
There's a hierarchy for that test now. What do you think?

fniephaus:
Yes, LGTM, thank you! When can this be merged? There is still a job failure.

theseion:
#607 needs to be merged first.

fniephaus:
Can you cherry-pick the changes so we see a green gate?

theseion:
Done.

fniephaus (approved):
LGTM, thanks!

PR #607, opened 2023-09-24
feat: update Pharo VM and image urls

theseion:
Note: This commit also changes `Pharo-stable` and `Pharo-alpha` to be synonyms of `Pharo64-stable` and `Pharo64-alpha`, respectively. Previously, these aliases referred to the 32-bit versions.


theseion:
See discussion in https://github.com/hpi-swa/smalltalkCI/pull/606.

theseion:
Unfortunately, the Pharo deprecation stuff is in #606, so the tests here will never become green.

PR #621, opened 2023-12-23
feat: always update development images

theseion:
- new option to force download of an image and override the cache
- never use the cached image for in-development versions (*-trunk | *-alpha)

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -286,12 +286,12 @@ pharo::prepare_image() {
     mkdir "${target}"
   fi
   if ! is_file "${target}"/*.image; then
-    pushd "${target}" >/dev/null
+    pushd "${target}" > /dev/null
     fold_start download_image "Downloading ${smalltalk_name} image..."
       download_file "${pharo_image_url}" "${pharo_zeroconf}"
       retry 3 "bash ${pharo_zeroconf}"
     fold_end download_image
-    popd >/dev/null
+    popd > /dev/null
   fi
 
   print_info "Preparing Pharo image..."
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -278,7 +278,7 @@ pharo::prepare_image() {
   local target="${SMALLTALK_CI_CACHE}/${smalltalk_name}"
   local pharo_zeroconf="${target}/zeroconfig"
 
-  if "${config_update_image}" && is_dir "${target}"; then
+  if "${config_overwrite_cache}" && is_dir "${target}"; then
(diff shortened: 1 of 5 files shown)


theseion:
Note: this change is necessary due to a fix a couple of months back that fixed the caching behaviour (#604).

run.sh
```diff
@@ -344,6 +360,10 @@ parse_options() {
       fi
       shift 2
       ;;
+    -u | --update-image)
```
fniephaus, line 363: `--update-image` sounds like it triggers an in-image update. Maybe something like `--refresh-cache` or `--overwrite-cache`?

Now that I think about this, maybe trunk/alpha images should never be cached, or always overwritten. For dev purposes, we could also provide the inverse option, something like `--force-cache`.

BTW: we also already have `--clean` for cleaning the cache.

theseion: Yes, those sound better.

> Now that I think about this, maybe trunk/alpha images should never be cached, or always overwritten. For dev purposes, we could also provide the inverse option, something like `--force-cache`.

That was the main intent of this option. `--clean` clears the entire cache, I simply wanted to use the latest image for trunk / alpha.

pharo/run.sh
```diff
@@ -278,14 +278,20 @@ pharo::prepare_image() {
...
-    pushd "${target}" > /dev/null
+  fi
+  if ! is_file "${target}"/*.image; then
+    pushd "${target}" >/dev/null
     fold_start download_image "Downloading ${smalltalk_name} image..."
       download_file "${pharo_image_url}" "${pharo_zeroconf}"
       retry 3 "bash ${pharo_zeroconf}"
     fold_end download_image
-    popd > /dev/null
+    popd >/dev/null
```
fniephaus, line 294:
```suggestion
    popd > /dev/null
```

run.sh
```diff
@@ -280,12 +280,34 @@ validate_configuration() {
...
 
+################################################################################
+# Set options that depend on the contex, i.e., the input values and selections
+# performed so far.
+################################################################################
+set_context_options() {
+    if [ "${config_force_cache}" = "true" ]; then
+      print_info "Forcing cache use"
+      config_overwrite_cache="false"
+      return
```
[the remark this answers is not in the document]
theseion: Like this?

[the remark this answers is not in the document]
fniephaus: Yes, thanks!

fniephaus (approved):
LGTM

Issue #623, opened 2024-01-14
Pharo12 builds fail because they removed Time>>millisecondsToRun:


jbrichau:
The builds using Pharo12 fail because they removed the `Time>>millisecondsToRun:` method:

`Instance of Time class did not understand #millisecondsToRun:`

For example: https://github.com/SeasideSt/Seaside/actions/runs/7512588387/job/20453435450

jbrichau:
I fixed the mistake in the baseline for Pharo 3 that made builds fail after the class package change. I do wonder if these old versions should still be supported in the future though.

theseion:
This issue can be closed, right?

jbrichau:
Yes, thanks @theseion and @fniephaus !

PR #624, opened 2024-01-14
Fix Pharo12 builds failing due to removal of `Time>>millisecondsToRun:`

jbrichau:
Pharo12 builds started failing (see issue #623).

Changes in this PR fix this issue as shown in build https://github.com/SeasideSt/Grease/actions/runs/7518132269/job/20465637446
that uses the fork of this PR.

While making these changes, I also encountered a load dependency failure due to a class for the code coverage implementation in Pharo that was in the Core package rather than the Coverage package.

I also notice that because of the use of Iceberg, the (obsolete) monticello metadata has been removed. This should be fine especially for Pharo but let me know if you want things to be different.


jbrichau:
Ok, I removed metadata also from non-Pharo-specific packages and the builds require very old versions of Pharo to still work. I will re-create this PR.

PR #625, opened 2024-01-14
Fix Pharo12 builds failing due to removal of Time>>millisecondsToRun:

jbrichau:
Pharo12 builds started failing (see issue https://github.com/hpi-swa/smalltalkCI/issues/623).

Changes in this PR fix this issue as shown in build https://github.com/SeasideSt/Grease/actions/runs/7518460089
that uses the fork of this PR.

While making these changes, I also encountered a load dependency failure due to a class for the code coverage implementation in Pharo that was in the Core package rather than the Coverage package. I moved it to fix that issue as well.

(re-create of PR #624)


jbrichau:
I fixed the mistake in the baseline for Pharo 3 that made builds fail after the class package change. I do wonder if these old versions should still be supported in the future though.

fniephaus:
Thanks for fixing this, @jbrichau. @theseion, could you please take a look before we merge this? thanks!

> I do wonder if these old versions should still be supported in the future though.

If the Pharo community doesn't use them anymore, we can drop them. Users could still use an old version of smalltalkCI if they really need to test old versions. But as long as the changes are relatively small, I think we could keep them around.

theseion:
Something's fishy. They simply moved the classes to a different package. That method shouldn't have gotten lost. It appears that the build version is stuck at a commit from December.

theseion:
https://github.com/pharo-project/pharo/issues/15953#issuecomment-1892735684

theseion (commented):
The baseline changes fix the dependency issue when loading the baseline, I presume?

jbrichau:
@theseion The dependency issue is solved by moving the class between packages in commit 5ca88727574e32789afc6b3745ca86b0d248e079. But then I needed to also fix the baseline for pharo 3 and 4 because it was expecting the class to be in that wrong package (see https://github.com/hpi-swa/smalltalkCI/actions/runs/7518470953/job/20465859552)

theseion:
I should have read your description properly, sorry.

theseion (approved):

fniephaus (approved):
LGTM, thanks a lot @jbrichau!

PR #629, opened 2024-02-08
Updates for Pharo12 deprecated methods.

hernanmd:
This PR introduces two small changes to use updated versions of deprecated methods in Pharo 12, avoiding Debugger windows when running in interactive mode.

- Use `tags` instead of `classTags` (deprecated in Pharo12)
- Use `Author fullName:`  instead of `setAuthor` (deprecated in Pharo12).
- Changes in .json files are just a re-ordering of properties.

### DIFF
--- a/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo12.class/class/allCategoryNames.st
+++ b/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo12.class/class/allCategoryNames.st
@@ -7,5 +7,5 @@ allCategoryNames
 		streamContents: [ :stream |
 			packages do: [ :package |
 				stream nextPut: package name asString.
-				package classTags do: [ :tag |
+				package tags do: [ :tag |
 					stream nextPut: tag categoryName asString ] ] ]) asSet asArray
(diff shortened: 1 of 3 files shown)


fniephaus:
Thanks for the PR, @hernanmd. Any way we could clean up the diff and make sure only relevant changes are in the PR. That'd be great and simplify reviewing. Thanks!

And oh no, another set of deprecations in Pharo-12. Could you please take a look at this @theseion?

theseion (approved):

theseion:
Thanks @hernanmd. I've cleaned up the PR. @fniephaus please review and merge.

@fniephaus I fear at some point we'll see a breaking change in the VCS format (e.g., the `.categories` file). I don't know how to deal with that. For Fuel, I've completely abandoned compatibility and have a dedicated branch for Pharo12. I don't think that's a good option for SmalltalkCI though. Maybe we could convince the Pharo devs to ensure that VCS files remain compatible with previous releases.

fniephaus:
> I've cleaned up the PR. @fniephaus please review and merge.

Thanks a lot, Max! Next drink is on me ;)

> Maybe we could convince the Pharo devs to ensure that VCS files remain compatible with previous releases.

That'd be good. Some level of backward compatibility is required to allow a consistent, multi-version CI experience. Alternatively, we could consider going back to the fileIn format, unless Pharo devs have broken or will break that as well.

fniephaus (approved):
LGTM, thank you both!

theseion:
> That'd be good. Some level of backward compatibility is required to allow a consistent, multi-version CI experience. Alternatively, we could consider going back to the fileIn format, unless Pharo devs have broken or will break that as well.

I'll bring it up when I get the chance.

PR #634, opened 2024-02-15
adjusted saveImage for Pharo11 as it returns a SnapshotOperation instead

### LINKED ISSUE #631: SmalltalkCI class>>saveAndQuitImage currently broken in  Pharo-11.0.0+build.725
`self platformClass saveImage` is not returning a boolean so this method fails:

```smalltalk
saveAndQuitImage
	self platformClass saveImage
		"Close image only if it is not resuming."
		ifFalse: [ self platformClass quitImage ] 
```

 Pharo-11.0.0+build.725

Screen Shot 2024-02-15 at 11 37 53

I had a lot of succesful builds in Pharo 11 for several projects so I wonder why it broke now if it was previously working. What moved? Aren't versions used for this pinned?

noha:
of boolean

Fixes #631

### DIFF
--- a/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo11.class/class/saveImage.st
+++ b/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo11.class/class/saveImage.st
@@ -1,0 +1,3 @@
+overrides
+saveImage
+	^ super saveImage isImageStarting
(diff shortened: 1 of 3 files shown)


fniephaus (approved):
LGTM, thanks for the fix, @noha!

labordep:
Thanks for the fix! Looks good for Pharo 11👌

fniephaus:
Thanks for confirming, @labordep!

labordep:
@fniephaus I cannot see any changes for Pharo 12 CI (always the same issue), is it normal?

fniephaus:
The Pharo 12 compatibility class [inherits from the one for Pharo 11](https://github.com/hpi-swa/smalltalkCI/blob/adf5b117b44b426b631ef4f90a00eca7dffffa30/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo12.class/properties.json#L13), so this fix should apply to Pharo 11 and 12.

labordep:
Ok, this is another issue, I will open it.

PR #630, opened 2024-02-09
fix: change directory for Coveralls report

theseion:
In order to use `git` commands for generating the Coveralls report, the script needs to `cd` to `SMALLTALK_CI_HOME` first.

Fixes #626

PR #636, opened 2024-02-19
fix: get coveralls data from correct directory

theseion:
Refs #630

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -425,7 +425,6 @@ ensure_jq_binary() {
 
 git_log() {
   local format_value=$1
-  local output
   echo "$(git --no-pager log -1 --pretty=format:"${format_value}" | "${jq_binary}" -Rs .)"
 }
 
@@ -441,6 +440,14 @@ export_coveralls_data() {
   local service_name=""
   local service_pull_request=""
   local url="unknown"
+
+  # Change directory to the target project repository.
+  # If input was read from stdin, presume we're in the right location already 
+  if is_file "${config_ston}"; then
+    pushd "$(dirname ${config_ston})" > /dev/null
+  else
+    pushd "$(pwd)"
+  fi
 
   if ! grep -q "#coverage" "${config_ston}"; then
     return 0 # Coverage data not needed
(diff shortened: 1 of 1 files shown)


theseion:
Tested from within a repository, from outside a repository, and from outside a repository with renamed `.git` (no Git data).

helpers.sh
```diff
@@ -498,18 +509,31 @@ export_coveralls_data() {
 
   ensure_jq_binary # required for git_log
 
+
+  if ! git --no-pager log -1 > /dev/null 2>&1; then
```
theseion, line 513: FYI: tried to use `git_log` here but that doesn't report the exit code, even with `set -o pipefail`.

theseion:
Ping @fniephaus.

helpers.sh
```diff
@@ -442,6 +441,18 @@ export_coveralls_data() {
   local service_pull_request=""
   local url="unknown"
 
+  # Change directory to the target project repository.
+  # The assumption here is that the STON file resides within the
+  # repository.
+  # Otherise, don't change directory and presume that we're where
+  # we need to be.
+  if is_file "${config_ston}"; then
```
fniephaus, line 449: I wonder: should the current working directory take precedence over all of this? We could do it the other way around: if we are in a git repo, we stay there. otherwise, we try to switch to the directory of the ston. if that does not work, we skip this.

helpers.sh
```diff
@@ -498,18 +509,31 @@ export_coveralls_data() {
...
+    popd > /dev/null
+    return 0
+  fi
+
+  local author_email="$(git_log "%ae")"
+  local author_name="$(git_log "%aN")"
+  local committer_email="$(git_log "%ce")"
+  local committer_name="$(git_log "%cN")"
+  local commit_id="$(git_log "%H")"
+  local commit_message="$(git_log "%s")"
```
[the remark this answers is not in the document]
theseion: Yes, but given that we check before anyway, we don't need this. I had done it originally when I wanted to check the exit code after the assignment.

helpers.sh
theseion: Sounds reasonable. I'll change that.

helpers.sh
theseion: If we leave the check like this, no. But if I wanted to write `if ! git_log "%ae"; then...` it wouldn't work. That's why this is an extra line instead of checking the first use of `git_log` below.

fniephaus (approved):
LGTM

PR #638, opened 2024-03-29
Fix retrieval of package name for a method in Pharo 12

jbrichau:
This PR fixes a crash in Pharo 12 when code coverage reporting is enabled and when methods of a Trait are passed through `SCIPharo12CodeCoverage class>>#packageNameForMethod:`

Without the fix: https://github.com/SeasideSt/Seaside/actions/runs/8442597789/job/23124347278
With the fix: https://github.com/SeasideSt/Seaside/actions/runs/8482174261

I also notice metadata is removed. Since the changes are in a Pharo-specific package, I left them in. If you want me to revert those changes and keep metadata, let me know.


jbrichau:
Since removing the method metadata breaks loading for platforms as old as Pharo 3 (even though the package is not for Pharo 3), I reverted the metadata changes.

I would recommend to drop those old Pharo versions for the future ;-)

jbrichau:
The build failed because of a download failure. I cannot restart the build.

jbrichau:
@fniephaus I cannot re-run the action jobs on this PR. Could you? All the runs should be green again but the previous built failed because of a download failure.

fniephaus (approved):
LGTM, simpler is always better!

PR #643, opened 2024-05-21
feat: add PharoXX-13 image selection

theseion:


fniephaus (approved):
LGTM.

theseion:
@dalehenrich GemStone tests are failing, `#testSentButNotImplemented`. I suppose it's the use of `#at:ifAbsent:` that I introduced. Any idea how to handle this?

estebanlm (approved):
looks fine, thanks :)
and as I said... if this is staled is a problem on Pharo-13 that will also be there in Pharo-alpha so... we need to work on making it not-stale :P

fniephaus:
> I suppose it's the use of #at:ifAbsent: that I introduced. Any idea how to handle this?

Avoid the use of `#at:ifAbsent:` and use `#at:` and `#iNil:` instead?

dalehenrich:
Sorry, just noticed this. According to the log, it looks like there is a method #skip introduced introduced in SCIGoferLoadSpecTest>>testAddLoadedClassesFrom ... found this by looking in the log for the string `Sent but not implemented`:
```
2024-05-21T19:13:30.0518150Z .'Sent but not implemented from #''SCIGoferLoadSpecTest>>testAddLoadedClassesFrom'''
2024-05-21T19:13:30.0519082Z 	'#''skip'''
```
So it isn't `at:ifAbsent:` that's the issue ... presumably this is in code that is not used by GemStone? #perform: could be used ...

Here's the source and it seems that #skip is expected to be implemented in a superclass or ???
```
testing
testAddLoadedClassesFrom
	| goferClass gofer |
	goferClass := Smalltalk
		at: #Gofer
		ifAbsent: [ ^ self skip ].
	
	self assert: true equals: (self spec loadedClasses isEmpty).
	
	gofer := goferClass new
		package: 'SmalltalkCI-Core';
		package: 'SmalltalkCI-Tests'.
	self spec addLoadedClassesFrom: gofer references.
	self deny: (self spec loadedClasses isEmpty).
	self assert: (self spec loadedClasses includes: SmalltalkCI).
	self assert: (self spec loadedClasses includes: SmalltalkCITest).
```
... hmmm, Gofer is present in GemStone, so the test must pass for GemStone ...

theseion:
Thanks @dalehenrich. I added the `#skip` to skip the test for Pharo 13, since it no longer has `Gofer`. Is there a similar message I could send for GemStone to skip the test?

> Avoid the use of #at:ifAbsent: and use #at: and #iNil: instead?

Yes, I tried that, but `SystemDictionary` in Pharo 13 apparently no longer understands `#at:`.

dalehenrich:
@theseion 
> Thanks @dalehenrich. I added the `#skip` to skip the test for Pharo 13, since it no longer has `Gofer`. Is there a similar message I could send for GemStone to skip the test?

No ... an early `^self` will "skip the test" ... I assume that Pharo sunit must keep track of skipped tests?

theseion:
I see. Since this test is internal, I'll just use `^ self`.

PR #646, opened 2024-06-15
feat: report image info without `StSystemReporter`

theseion:
`StSystemReporter` was removed in Pharo 13.

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -70,7 +70,7 @@ print_help() {
     --headful               Open vm in headful mode and do not close image.
     --image                 Custom image for build (Squeak/Pharo).
     --install               Install symlink to this smalltalkCI instance.
-    --overwrite-cache   Download the newest image and cache it.
+    --overwrite-cache       Download the newest image and cache it.
     --print-env             Print all environment variables used by smalltalkCI
     --no-color              Disable colored output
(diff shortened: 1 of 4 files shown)


theseion:
@estebanlm, @guillep Here's such an instance where  the behavior of Pharo64-13 differs from Pharo64-alpha: https://github.com/hpi-swa/smalltalkCI/actions/runs/9525945437/job/26260957214?pr=646#step:4:236.

The method simply checks wether generating two random values twice produces a different identifier, equivalent of: 

```smalltalk
(ByteArray with: 255 atRandom with: 255 atRandom) hex ~= (ByteArray with: 255 atRandom with: 255 atRandom)
```

For some very weird reason, `Random>>#nextInteger:` produces the same value multiple times. At first, I thought it was a timing issue, so I fixed the use of `Random` in `#newTravisID`, but the test still fails. Interestingly, the test only fails on the CI, not on my machine. I suspect, that it's an issue that depends on OS and VM combination (can't test that on my machine).

demarey:
The Random problem in Pharo is now fixed: https://github.com/pharo-project/pharo/pull/16780.
I also proposed a PR to fi fix the problem of the system reporter in P13: https://github.com/hpi-swa/smalltalkCI/pull/649/

theseion:
@fniephaus good to go

fniephaus (approved):
LGTM, thanks!

PR #649, opened 2024-06-17
fix undeclared for Pharo 13

demarey:


theseion:
Already fixed in #646.

PR #657, opened 2024-08-27
Moose image extraction: add a flag to unzip only files

ClotildeToullec:
Moose images are now in a zipped directory. The `-j` flag ensures that only the files are unzipped.


fniephaus:
Thanks for the PR. Is this true for all supported versions of Moose?

theseion:
Looks like this fixes the issue in the other PR. I'll merge this one so we can get on with the other one.

ClotildeToullec:
It works for Moose 8 and above. 
Actually, I should make a PR to remove older versions, they are not available anymore.

PR #659, opened 2024-08-29
Remove unsupported Moose images

ClotildeToullec:
Removing references to old versions of Moose that are no longer available.


theseion (approved):
Thanks!

PR #662, opened 2024-09-17
feat: only instantiate test resources once in Pharo

### LINKED ISSUE #661: Test resources are set up multiple times
Given a TestResource subclass DummyTestResource with `#setUp` implemented as showing 'Setting up DummyTestResource' on the transcript, after adding DummyTestResource to the resources for SCIAbstractLoadSpecTest, the message is shown five times, rather than just once, when running the smalltalkCI tests using Pharo64-13.

Reverting the change made to `#runClasses:spec:` for SCIPharoTestRunner in commit 7fed37997820de11 causes the message to only be shown once as expected (but also causes the test `#testCustomTestSuite` introduced in the same commit to fail).

theseion:
Fixes #661

Note: also includes a change to the project loading code. The Moose images already contain SmalltalkCI, which prevented the changes in this PR from being picked up. We now force load SmalltalkCI in Pharo and GToolkit builds.


Rinzwind:
Looks OK. A slight improvement that could be made is to make `#testResourcesSetupOnce` assert that the counters are nil before the send of `#runSpec:` so as to make the change to the counters clear in the test. The exclusion of SCIPharoTestSuiteNestingTestCase1 and SCIPharoTestSuiteNestingTestCase2 could perhaps be handled by adding them to the `#exclude` in ‘.smalltalk.ston’ instead of through the reflective `#isAbstract` implementation.

theseion:
> Looks OK. A slight improvement that could be made is to make #testResourcesSetupOnce assert that the counters are nil before the send of #runSpec: so as to make the change to the counters clear in the test. 

Good idea.

> The exclusion of SCIPharoTestSuiteNestingTestCase1 and SCIPharoTestSuiteNestingTestCase2 could perhaps be handled by adding them to the #exclude in ‘.smalltalk.ston’ instead of through the reflective #isAbstract implementation.

I chose `#isAbstract` on purpose, to prevent these tests from being available for running in the image as well. They would fail and that would be difficult to understand.

theseion:
Ping @fniephaus

fniephaus (approved):
LGTM, although I don't have enough time to go through every detail of this change at the moment.

PR #665, opened 2024-12-19
Rename env variable for GitLab build

badetitou:
The environment variable has been rename on GitLab. Deprecated in version 9.x and removed now.

I think we can replace them in the script now.
It creates bugs 

see https://github.com/fsprojects/FAKE/issues/2290#issuecomment-522197031

### DIFF
--- a/helpers.sh
+++ b/helpers.sh
@@ -490,8 +490,8 @@ export_coveralls_data() {
     url="https://github.com/${APPVEYOR_REPO_NAME}.git"
   elif is_gitlabci_build; then
     branch_name="${CI_COMMIT_REF_NAME}"
-    service_job_id="${CI_BUILD_ID}"
-    service_job_number="${CI_BUILD_NAME}"
+    service_job_id="${CI_JOB_ID}"
+    service_job_number="${CI_JOB_NAME}"
     service_name="gitlab-ci"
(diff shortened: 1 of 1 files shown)


helpers.sh
```diff
@@ -490,8 +490,8 @@ export_coveralls_data() {
     url="https://github.com/${APPVEYOR_REPO_NAME}.git"
   elif is_gitlabci_build; then
     branch_name="${CI_COMMIT_REF_NAME}"
-    service_job_id="${CI_BUILD_ID}"
-    service_job_number="${CI_BUILD_NAME}"
+    service_job_id="${CI_JOB_ID}"
+    service_job_number="${CI_JOB_NAME}"
```
gcotelli, line 494: Maybe we can keep using the old name in case the new one is undefined? So it will not break the builds for people running its own Gitlab instance if it's not up to date.
```suggestion
    service_job_id="${CI_JOB_ID:-CI_BUILD_ID}"
    service_job_number="${CI_JOB_NAME:-CI_BUILD_NAME}"
```

gcotelli (changes requested):

gcotelli (approved):

badetitou:
Thanks

fniephaus:
Thank you for fixing!

PR #671, opened 2025-05-21
Add Moose 13

ClotildeToullec:
Moose 12 has been released as stable.
Moose 13 is the dev version. It runs on Pharo 13.

### DIFF
--- a/pharo/run.sh
+++ b/pharo/run.sh
@@ -109,7 +109,7 @@ moose::get_image_url() {
 
   case "${smalltalk_name}" in
     "Moose64-trunk"|"Moose-trunk")
-      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose12-development-Pharo64-12.zip"
+      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose13-development-Pharo64-13.zip"
       ;;
     "Moose64-8"*)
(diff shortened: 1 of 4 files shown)


theseion:
Thanks. I have to update the GitHub runner and then update this PR.

pharo/run.sh
```diff
@@ -124,7 +124,10 @@ moose::get_image_url() {
       echo "https://github.com/moosetechnology/Moose/releases/download/v11.x.x/Moose11-stable-Pharo64-11.zip"
       ;;
     "Moose64-12"*)
-      echo "https://github.com/moosetechnology/Moose/releases/download/continuous/Moose12-development-Pharo64-12.zip"
+      echo "https://github.com/moosetechnology/Moose/releases/download/v12.0.0/Moose12-stable-Pharo64-12.zip"
```
fniephaus, line 127: Will there ever be a 12.0.1 or so?

fniephaus (approved):
LGTM

pharo/run.sh
ClotildeToullec: Ah, yes, I should have used v12.x.x.
There will most probably not have a minor version update.
Moose major versions change approximately each year.
It is mostly a cluster of different projects and the finer versioning is managed at their level.
If by any chance we need a minor version change, I'll correct the version to v12.x.x.

PR #672, opened 2025-05-22
chore: update Linux runner on GitHub

theseion:
Ubuntu 20.04 has been deprecated

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-20.04, macos-latest, windows-2019 ]
+        os: [ ubuntu-latest, macos-latest, windows-2019 ]
         smalltalk:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-20.04, macos-latest, windows-2019 ]
+        os: [ ubuntu-24.04, macos-latest, windows-2019 ]
```
fniephaus, line 22: Should we be upgrading to ubuntu-22.04 (oldest compatible Ubuntu) or ubuntu-latest (consistent with macOS and automatically upgraded)?

theseion: I think, latest should be fine, but it could lead to some issues. I guess we'll see.

fniephaus:
Let's merge this now, it's already an improvement :)

PR #673, opened 2025-05-23
add Pharo14

estebanlm:
now yes, let's see how it goes :)

### DIFF
--- a/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo13.class/class/isPlatformCompatible.st
+++ b/repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo13.class/class/isPlatformCompatible.st
@@ -1,4 +1,4 @@
 compatibility
 isPlatformCompatible
 
-	^ self platformNameBeginsWith: 'Pharo13'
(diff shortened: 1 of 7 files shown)


estebanlm:
we need to wait for zeroconf to be there...

repository/SmalltalkCI-Pharo-Core.package/SmalltalkCIPharo13.class/class/isPlatformCompatible.st
theseion, line 1: Please create a subclass of `SmalltalkCIPharo13` instead, just like `SmalltalkCIPharo13` is a subclass of `SmalltalkCIPharo12`. You shouldn't need to implement any methods except for `isPlatformCompatible`.

theseion (changes requested):

estebanlm:
meh, is annoying because I need to do it by hand as the filetree format is not deterministic and generates a lot of garbage. 
I want to move all required classes into the image directly. 
That way stCI will just execute "as is" without loading anything... tbh for me will be easier to maintain :P

theseion:
😄 For you, yes, for us, not necessarily

theseion (approved):

PR #677, opened 2025-09-09
fixes #676 test runner command name for P14

### LINKED ISSUE #676: Update needed for latest P14
Hello,

We recently updated commands in Pharo to use Clap (Command line lib).
Smalltalk CI directly references the Pharo test command to compute the classes to be selected for testing.
The method is still there but the class name changed from `TestCommandLineHandler`to `ClapTestRunner`.

I would have been pleased to propose a PR but the fact that Samlltalk CI still uses a filetree format with metadata makes it impossible to do it from Pharo. Side question: what is the added value of using the metadata full format? Is a switch to another format planned?

Here is a patch with a very few changes that should be ok to use SmalltalkCI with P14. Let me know if it is ok or not.

Regards

[p14-test-command.patch](https://github.com/user-attachments/files/22229481/p14-test-command.patch)

demarey:


estebanlm (approved):

fniephaus:
Thank you!

PR #680, opened 2025-10-07
add Pharo14 compatibility method

estebanlm:
since we now add results in a "history" (in fact, this should have been always the case, it was just not working). 
and no need to actually register anything if executing on smalltalkCI, so this can remain empty.


estebanlm:
I added just one file @theseion  @fniephaus ... I have no idea what can be wrong :)

fniephaus (approved):
LGTM

estebanlm:
thanks !

PR #681, opened 2025-10-09
fix a call to a method deprecated in Pharo14

estebanlm:
as this will be deprecated then removed, I am adding a workaround (but keeping same functionality)


estebanlm:
this should be working @fniephaus ;)
sorry for the mess, but we are moving some things :P

PR #684, opened 2026-01-03
Add the #silently option of Metacello

jecisc:
SmalltalkCI is running mostly in CIs and updating the progress bars can take some time.

The load time can be reduced I think if we use Metacello silently

### DIFF
--- a/repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/basicLoadProjectOn..st
+++ b/repository/SmalltalkCI-Core.package/SCIMetacelloLoadSpec.class/instance/basicLoadProjectOn..st
@@ -7,6 +7,7 @@ basicLoadProjectOn: aSmalltalkCI
 	
 	metacello := (Smalltalk at: #Metacello) new.
 	metacello
+		silently;
 		repository: (self repository ifNil: [ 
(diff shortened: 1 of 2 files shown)


LinqLover:
Cool, did you measure the performance gain? This should maybe be disabled when running in headful mode?

jecisc:
I think it will depend on the Smalltalk running SmalltalkCI. It's possible that some will not have any difference. 

For SmalltalkCI I tried it on a project that is "middle size" and I gained 10-20sec on the run.

In an image headfull with multiple browsers open I have for the same project a load of 2.6sec that dropped at 0.6sec because most of the time is spend in the UI update when progress bars appears

LinqLover:
Squeak or Pharo? Preferably we should fix this UI overhead upstream anyway. At least in Squeak, progress bar updates tend to be cheaper than label changes (which may cause the progress bar widget to resize).

jecisc:
This is Pharo 13 and 14. In a Metacello loading the label tend to change a lot so it has a big impact (we have the same problem in Pharo about labels). In Pharo we will probably not change the progress bar for now since we will change the graphical stack. But we should check that the new one is faster :)

I almost never use Squeak

PR #689, opened 2026-07-18
pharo: update versions

theseion:
Also cleans up unsupported versions. Specifically, downloads for 64-bit VMs and images are only available starting with Pharo version 9.0.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,10 +37,6 @@ jobs:
           - Pharo64-11
           - Pharo64-10
           - Pharo64-9.0
-          - Pharo64-8.0
-          - Pharo64-7.0
-          - Pharo64-6.1
-          - Pharo64-6.0
           - Pharo32-stable
           - Pharo32-alpha
           - Pharo32-3.0
@@ -49,7 +45,6 @@ jobs:
           - Moose64-11
           - Moose64-10
           - Moose64-9.0
-          - Moose64-8.0
           - GemStone64-3.5.8
           - GemStone64-3.6.8
           - GemStone64-3.7.1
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -37,10 +37,6 @@ jobs:
           - Pharo64-11
           - Pharo64-10
           - Pharo64-9.0
-          - Pharo64-8.0
```
fniephaus, line 40: Why are we dropping all of these?

pharo/run.sh
```diff
@@ -39,24 +42,16 @@ pharo::get_image_url() {
...
-      echo "get.pharo.org/64/80"
-      ;;
-    "Pharo64-7.0")
-      echo "get.pharo.org/64/70"
-      ;;
-    "Pharo64-6.1")
-      echo "get.pharo.org/64/61"
-      ;;
-    "Pharo64-6.0")
-      echo "get.pharo.org/64/60"
```
fniephaus, line 52: Is there a good reason to break projects that rely on 6 etc?

.github/workflows/main.yml
theseion: Because those images and VMs no longer exist.

pharo/run.sh
theseion: Those images and VMs are no longer available.

fniephaus: Thanks, I was not aware!

theseion: Me neither. I only noticed when the tests ran that those resources are gone.

fniephaus: Too bad someone decided to just delete these, but ok. Thanks for cleaning this up in smalltalkCI!

PR #694, opened 2026-08-07
fix build issues

theseion:


theseion:
@fniephaus I've figured out that the older VMs for Squeak no longer work because we use newer Ubuntu versions (SqueakSSL requires dependencies there that cannot be satisfied on newer OS versions). Most builds are fine with just the newest VMs. However, I'm struggling to find a VM that works for Squeak32-4.5. Do you think it's worth the effort to keep support for Squeak32-4.5?

fniephaus:
Thanks for digging into this, @theseion!

> I'm struggling to find a VM that works for Squeak32-4.5. Do you think it's worth the effort to keep support for Squeak32-4.5?

 I'm not sure I can give a good answer. Maybe @marceltaeumel or @LinqLover have an opinion on this or knows what VM we could try? The latest 6.1 Squeak release seems to use [OSVM 202606270913](https://github.com/OpenSmalltalk/opensmalltalk-vm/releases/tag/202606270913). Maybe it makes sense to use it across all Squeak images if it's backward compatible. If we can't find a working VM, I'm ok with retiring support for Squeak32-4.5.

LinqLover:
> I've figured out that the older VMs for Squeak no longer work because we use newer Ubuntu versions (SqueakSSL requires dependencies there that cannot be satisfied on newer OS versions).

Is that related to this known issue? https://squeak.org/release_notes/6.0/#:~:text=Compatibility%20Notes-,SqueakSSL%20is%20not%20supported%20under%20Ubuntu%2016.04%20LTS.,-All%20code%20literals

> I'm ok with retiring support for Squeak32-4.5.

I found at least two actual repos on GitHub relying on that version for smalltalkCI, so I found that a bit sad.

Couldn't we just say (and document): if you want to use SqueakSSL, you either have to use a newer VM or an older version of Ubuntu in your runner (potentially in a docker image)?

theseion:
> Is that related to this known issue? https://squeak.org/release_notes/6.0/#:~:text=Compatibility%20Notes-,SqueakSSL%20is%20not%20supported%20under%20Ubuntu%2016.04%20LTS.,-All%20code%20literals

Not quite. SqueakSSL works on newer Ubuntu versions, just not as part of the older VMs, where it relies on libraries that aren't present on the system (but were on older Ubuntu versions).

>I found at least two actual repos on GitHub relying on that version for smalltalkCI, so I found that a bit sad.
>
>Couldn't we just say (and document): if you want to use SqueakSSL, you either have to use a newer VM or an older version of Ubuntu in your runner (potentially in a docker image)?

Good to know, thanks. We could just deprecate it for newer VMs. Unless we find a VM that can run 4.5 on a newer OS, which is what I was trying to figure out.

fniephaus:
> SqueakSSL works on newer Ubuntu versions, just not as part of the older VMs, where it relies on libraries that aren't present on the system (but were on older Ubuntu versions).

Is there any way we can install missing libraries on the newer systems? Maybe the bug is here: https://github.com/hpi-swa/setup-smalltalkCI/blob/b6a2d26f3cb2a4f2d89a99bf5658389c387c4c83/src/main.js#L20

fniephaus:
Also, does Squeak32-4.5 still run on `ubuntu-22.04`?

fniephaus:
> [!WARNING]
> OpenSSL 1.1.1 reached its official End of Life (EOL) in late 2023. While installing this older library is a necessary workaround to run legacy 32-bit applications, it will no longer receive standard security updates. You should avoid relying on it for modern or security-critical software.

It can still be installed on 24.04 and then the Squeak32-4.5 seem to work:
```
wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2.24_i386.deb
sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2.24_i386.deb
```

Should we add this (and a warning) to setup-smalltalkCI?

theseion:
That sounds reasonable. We can also add some documentation to the readme. By "setup-smalltalkCI" you mean the GH action, right?

fniephaus:
> We can also add some documentation to the readme. By "setup-smalltalkCI" you mean the GH action, right?

Yes, and because it's specific to GH Actions, I think a warning may be all we need.

fniephaus:
I added the libssl workaround to `setup-smalltalkCI` and got `Squeak32-4.5` to pass again. However, other Squeak images are still failing, I ran some experiments on [`fix-squeak32-builds`](https://github.com/hpi-swa/smalltalkCI/tree/fix-squeak32-builds). I will try to take another look soon.

marceltaeumel:
Current OSVM releases for 32-bit (squeak.cog.v3) work fine with a recent Windows container. In general, it makes sense to just use a recent Windows runner for older Squeak versions. No need to stop supporting older Squeak versions at this point.

grafik

LinqLover:
@fniephaus FWIW, @marceltaeumel and me have added support for further 32-bit VM plugins on recent Ubuntusv in https://github.com/LinqLover/setup-smalltalkCI/tree/marceltaeumel/install-path. It's on my list to create a PR for this :-)

